PR #23, opened 2026-03-03
Sharareh/unified evaluator

sharareh-y:
### What Changed

1. **Separated the evaluation process from each `launch_agent` into a unified evaluator.** The evaluator modules are:
   - `src/evaluator.py`: Centralized evaluation orchestrator providing compilation checking, correctness validation, and performance measurement
   - `src/evaluator_utils.py`: Utility functions for command execution and evaluation helpers
   - `src/performance.py`: Parses `execution_time` from different performance report files (handles different paths to support all task types)
   - `src/testcases.py`: Parses test cases and parameters
   - `src/plotting.py`: Generates per-test case performance speedup plots

2. **Updated `main.py` workflow** to use the unified evaluator:
   - First, the original kernel is compiled (required for baseline measurement in some task types)
   - Baseline performance is measured and stored in `baseline_perf.yaml` in the workspace
   - `launch_agent` is called for kernel optimization
   - `evaluate_kernel` is called to measure performance of the optimized kernel, stored in `optimized_perf.yaml` in the workspace
   - Per-test case performance speedup plots are generated
   - Task results (compilation, correctness, speedup, and score) are written to `task_result.yaml`

This centralizes evaluation logic, making it easier to maintain and ensuring consistent evaluation across all agents and task types.

### Testing

Tested the unified evaluator using the Cursor agent with `config.yaml` on 5 diverse tasks across different task types:
- `hip2hip/assign_score_withk`
- `hip2hip/silu`
- `torch2hip/gpumode/10099_Gather`
- `triton2triton/rocmbench/easy/test_add_kernel`
- `triton2triton/vllm/triton_apply_write`

Results (from `workspace_MI300_cursor/task_results_report.txt`):
- ✅ 100% compilation pass rate (5/5)
- ✅ 100% correctness pass rate (5/5)
- ✅ 60% speedup > 1.0 (3/5 tasks)
- ✅ Average speedup: 1.88x

This confirms the unified evaluator works correctly across different task types and successfully handles baseline measurement, optimized kernel evaluation, and result reporting.
performance_speedup

### DIFF
--- a/main.py
+++ b/main.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from src.tasks import get_task_config
 from src.preprocessing import setup_workspace, setup_rocm_env
 from src.module_registration import AgentType, load_agent_launcher, load_post_processing_handler
+from src.evaluator import measure_baseline, evaluate_kernel, write_task_result
 
 
 parser = argparse.ArgumentParser(description="arguments for AgentKernelArena")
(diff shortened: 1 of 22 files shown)


irvineoy:
Hi @sharareh-y, I noticed a small issue with the speedup calculation for multi-testcase tasks.

These two calculations are not equivalent.

Example:
- Case A: 100 ms -> 100 ms, speedup = 1x
- Case B: 1 ms -> 0.5 ms, speedup = 2x

If we want each testcase to have equal weight, we should average the per-testcase speedups:

- (1 + 2) / 2 = 1.5x

But if we instead divide the average baseline time by the average optimized time:
- average baseline = (100 + 1) / 2 = 50.5
- average optimized = (100 + 0.5) / 2 = 50.25
- speedup = 50.5 / 50.25 = 1.005x

So average(per-testcase speedup) and speedup from average times can produce very different results. If each testcase should count equally, we should use the first method consistently.

I pushed a small follow-up commit to fix this. Could you please take a look?

sharareh-y:
Hi @irvineoy, that's a good observation. Actually, the speedup reported in `task_result.yaml` is average per-test case speedup but the `score.py` used best_optimized_execution_time/base_execution_time for speedup calculation as before for scoring and ignored the speedup_ratio in `task_result.yaml`.
Your new commit correctly reads the speedup_ratio from `task_result.yaml` for score calculation.

PR #26, opened 2026-03-23
Features/add repo task

yueliu14:
Add repository-level task support

Path:
tasks/repository/*
src/preprocessing.py

Changes:

Add repo_url to config.yaml
Detect repo tasks through repo_url
Clone the repository if it does not exist, then copy it into the run workspace
If the repo directory exists but .git is missing, remove it and re-clone the repo

### DIFF
--- a/tasks/repository/rocprim/block_radix_rank/config.yaml
+++ b/tasks/repository/rocprim/block_radix_rank/config.yaml
@@ -1,4 +1,9 @@
 repo_url: https://github.com/ROCm/rocPRIM.git
+task_type: repository
+repository_language: hip
+post_clone_install:
+  - command -v cmake >/dev/null 2>&1 || (apt-get update && apt-get install -y cmake ninja-build)
+post_clone_install_mode: every_setup
 compile_command:
   - python3 scripts/task_runner.py compile
 correctness_command:
--- a/tasks/repository/rocprim/block_radix_rank/config.yaml
+++ b/tasks/repository/rocprim/block_radix_rank/config.yaml
@@ -1,0 +1,10 @@
+repo_url: https://github.com/ROCm/rocPRIM.git
+compile_command:
+  - python3 scripts/task_runner.py compile
+correctness_command:
+  - python3 scripts/task_runner.py correctness
+performance_command:
+  - python3 scripts/task_runner.py performance
+prompt:
+  cheatsheet: null
+  instructions: "Optimize block_radix_rank"
--- a/tasks/repository/rocprim/block_radix_rank/scripts/task_runner.py
+++ b/tasks/repository/rocprim/block_radix_rank/scripts/task_runner.py
@@ -1,0 +1,307 @@
+#!/usr/bin/env python3
+# Copyright(C) [2026] Advanced Micro Devices, Inc. All rights reserved.
+"""
+Task runner for repository/rocprim/block_radix_rank.
+
+This script provides a stable interface for AgentKernelArena's evaluator:
+  - `compile`      : configure & build rocPRIM benchmark/test targets
+  - `correctness`  : run `test_block_radix_rank`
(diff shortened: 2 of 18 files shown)


tasks/repository/rocprim/block_radix_rank/config.yaml
```diff
@@ -0,0 +1,10 @@
+repo_url: https://github.com/ROCm/rocPRIM.git
```
irvineoy, line 1: These new repository task configs don't set `task_type`, but the normal agent flow goes through `src/prompt_builder.py`, which raises on missing `task_type` and only knows the existing task families. As soon as one of these tasks is selected, `cursor`/`codex`/etc. will fail before the agent even starts. We need to either add a supported task type here and teach the prompt builder/cheatsheets about it, or route repository tasks through a different prompt path.

tasks/repository/rocprim/block_radix_rank/scripts/task_runner.py
```diff
@@ -0,0 +1,307 @@
...
+        report = {"status": "ok" if ok else "fail", "error": err}
+        (report_root / "correctness_report.json").write_text(json.dumps(report, indent=2))
+        print(f"Correctness: {'PASS' if ok else 'FAIL'}")
+        if err:
+            print(err)
+        sys.exit(0 if ok else 1)
+
+    elif args.mode == "performance":
+        results, err = run_performance(ws, trials=args.trials)
+        (report_root / "performance_report.json").write_text(json.dumps(results or [], indent=2))
```
irvineoy, line 293: The runner writes throughput (`bytes_per_second_gs`) to `performance_report.json` and prints `G/s` to stdout, but the evaluator only parses execution-time fields like `execution_time_ms`/`time_ms` (or `ori_time`/`opt_time` for torch2hip). In practice `measure_baseline()`/`measure_performance()` will return no valid test cases for these tasks, so speedup/scoring become zero even if the benchmark ran. This needs to emit execution time in the existing schema, or the evaluator needs a rocPRIM-specific parser.

irvineoy (changes requested):
I found two blocking issues that need to be resolved before these repository tasks can run correctly in the current Arena pipeline.

sharareh-y:
These fixes are applied to the PR: 
### Added `task_type: repository`
- Added **`repository_task_type()`** in `src/prompts/task_type.py` (full-repo / HIP or Triton scope, not single-file hip2hip).
- **`src/prompt_builder.py`**: dispatches `repository`, uses **`repository_language`** (`hip` | `triton`) to pick the knowledge cheatsheet from `default_cheatsheet.yaml`.
---

### Post-clone system dependencies

- **`src/preprocessing.py`**: after `git clone`, optional **`post_clone_install`** (list of shell commands) and **`post_clone_install_mode`**: `after_clone` (default) vs `every_setup` (e.g. guarded `apt-get` when the repo already exists but the container is new).
- RocPRIM example tasks include installing **cmake** / **ninja** where appropriate.

---

### Evaluator timeouts

- **Compile**: default **3600s** per command; optional **`compile_timeout`** in task `config.yaml`.
- **Correctness**: default **3600s**; optional **`correctness_timeout`** (addresses rocPRIM correctness runs hitting the old **300s** limit).
- **Performance**: **`measure_performance`** no longer uses a fixed **300s**; uses **`performance_timeout`** (default **3600s**).

---

### Performance reporting (`bytes_per_second_gs`)

- Repository benchmarks (e.g. rocPRIM) emit **throughput** (`bytes_per_second_gs`), not wall time. The evaluator previously only recognized time-like keys, so **0 test cases** were parsed and baseline/optimized stayed empty.
- **`src/testcases.py`**: if `bytes_per_second_gs` is present, use **cost = 1/throughput** so existing speedup math **`baseline_cost / optimized_cost`** equals **opt_throughput / baseline_throughput**.

PR #27, opened 2026-03-30
[FIX] chaned loaded kernel from .cu to .hip

chushic:
fix cases: gather_points and roiaware_pool3d


irvineoy:
Good catch, thanks a lot!

PR #28, opened 2026-04-06
Geak triton common benchmark

iraj465:
Adds triton tasks to Arena.


tasks/triton2triton/geak_eval/L1/fused_append_shared_experts/test_kernel_harness.py
sharareh-y, line 1: Blocking. This harness is hardcoded to load `fused_moe_triton_kernels.py`, but the task config (`source_file_path: kernel.py`) and both launchers (`geak_v3_triton`, `mini_swe_triton`) only edit `kernel.py`. After the agent runs, the two files diverge and the harness keeps importing the unmodified duplicate — so `--correctness` always passes and `--full-benchmark` always reports the baseline. Effective speedup will be 1.0× regardless of the agent.

Two options:
  a) Drop `fused_moe_triton_kernels.py` (it's a byte-identical 977-line copy of `kernel.py`) and change `_KERNEL_FILENAME` to `kernel.py`. [Prefered]
  b) Keep the sglang-style filename and update `source_file_path` + the launchers to operate on `fused_moe_triton_kernels.py`.

Either is fine, but the current setup is broken end-to-end.

scripts/run_geak_triton.sh
sharareh-y, line 1: Blocking. Please remove the user-specific fallbacks:
  CONTAINER="geak-agent-${USER:-sapmajum}"
  export GEAK_SRC="${GEAK_SRC:-/home/sapmajum/GEAK-agent-filtering-and-cli-unification/src}"

Both leak one developer's environment into shared tooling. Either hard-fail when GEAK_SRC / $USER isn't set, or document the required env vars at the top and use `: "${GEAK_SRC:?GEAK_SRC must be set}"`.

Also: the .tmp_config_stream_{a,b}.yaml temp files are written into the repo root and only cleaned up on the happy path. With `set -euo pipefail` and the `|| FAIL=1` swallow on `wait`, a crash leaves dirty files in the worktree. Please wrap cleanup in a `trap`.

src/evaluator.py
```diff
@@ -217,7 +402,28 @@ def evaluate_kernel(
...
+            results['geak_best_round'] = geak_results.get('best_round')
+            results['geak_round_history'] = geak_results.get('round_history', [])
+            log.info(
+                f"Using GEAK verified results: {geak_results['verified_speedup']:.4f}x "
+                f"(baseline={geak_results['baseline_ms']:.4f}ms, "
+                f"candidate={geak_results['candidate_ms']:.4f}ms, "
+                f"benchmark={geak_results.get('benchmark_speedup', 'N/A')}x, "
+                f"task={geak_results.get('best_task', 'N/A')})"
+            )
+
```
sharareh-y: Blocking.
Please remove this part, we want final evaluation to come from AgentKernelArena.

agents/geak_v3_triton/README.md
sharareh-y, line 1: This README is out of sync with the code it's documenting:
  - Defaults table says GEAK_MAX_ROUNDS=5 and GEAK_MODEL_ENSEMBLE=gpt-5.2,claude-opus-4.6, but agent_config.yaml ships MAX_ROUNDS=3 and has no MODEL_ENSEMBLE set.
  - "All 18 Triton Kernels" table lists ff_backward at L1, but config_geak_triton_all18.yaml lists it at L2.
  - Same table has a `rope` kernel at L2 (# 11) — there's no rope task in config_geak_triton_all18.yaml. Did you mean fused_qk_rope_cache_mla / fused_qkv_rope?
  - L1 row count is 7 in the README, 6 in the config.

sharareh-y (changes requested):
Thanks for your contribution, overall code looks good, here I things I'd like addressed:

1. fused_append_shared_experts is wired so the agent edits kernel.py but the harness only loads fused_moe_triton_kernels.py — see file-level comment. Speedup for this kernel will always be ~1.0×.
2. User-specific hardcoded paths are present in code: `sapmajum` username and `/home/sapmajum/...` GEAK_SRC fallback in scripts/run_geak_triton.sh, `/workspace/src` in agents/mini_swe_triton/launch_agent.py, `/sgl-workspace/aiter` default in src/evaluator_utils.py.
3. In src/evaluator.py GEAK results overwrites report if AKA fails. We don't want this behavior, final evaluation for all agents should come from AKA runs.
4. agents/geak_v3_triton/README.md and the agent_config.yaml disagree on defaults (MAX_ROUNDS 5 vs 3, MODEL_ENSEMBLE listed but not set) and the "All 18 kernels" table doesn't match config_geak_triton_all18.yaml (ff_backward listed at L1, mystery `rope` row, etc.).

sharareh-y:
Thanks for this. I went through the branch and, to get it running end-to-end, I brought it up to date with `main` and pushed a few fixes on top.
Everything below is now working with **AgentKernelArena as the single source of truth** for scoring: I verified on 8× MI308X (gfx942) inside the sglang ROCm container that `geak_v3_triton` (e.g. `triton_rms_norm` → ~1.14×) and `geak_v3` (HIP `knn`) both complete baseline → GEAK optimize → AKA re-evaluate, and I ran the task_validator across the `geak_eval` suite.

## What I fixed / added

- **Merged `main` (docker-first).** Pulled in main's docker-first workflow (`make docker-*` / `src/scripts/docker_benchmark.sh`), parallel execution in `main.py`, perf-helper materialization, `task_validator`, and the tasks/docs reorg. I resolved the conflicts so AKA re-evaluates independently (no GEAK  `final_report` fallback) — GEAK auto-applies its winning patch and AKA re-measures it.
- **Restored Triton stdout latency parsing.** On main, `parse_execution_time` had its stdout strategy commented out and relied only on JSON reports. Your Triton `geak_eval` harnesses emit latency solely on stdout as
  `GEAK_RESULT_LATENCY_MS`, so every stdout-only task was scoring `0.00×`. I re-enabled the stdout fallback (JSON tasks still resolve first).
- **Pruned the agent registry.** Removed enum members / imports / handlers for the agents that were deleted on main (`geak_hip`, `geak_optimagentv2`, `openevolve`, `single_llm_call`, `geak_ourllm_kernel2kernel`, `swe_agent`).
- **geak_eval task fixes.**
  - Removed `ff_backward` — the kernel crashes the Triton `make_ttgir` pass with an LLVM `AMDMfmaEncodingAttr` assertion (`PassManager::run failed`) on gfx942 / triton 3.6.
  - Fixed `fused_qk_rope_cache_mla` `target_kernel_functions` (it named a symbol that doesn't exist; correctness itself already passed 128/128).
- **Removed `sitecustomize.py`** — the namespace-stub hack only ran in the orchestrator process, wasn't doing anything for the runs, and its directory-renaming side-effect was risky.
- **Config cleanup.** Restored main's `config.yaml` (template `cursor` + the standard task set) and added `geak_v3` / `geak_v3_triton` to the templates comment; removed the ad-hoc run-configs and the now-orphaned
  `scripts/run_geak_triton.sh`.
- **Docs.** Rewrote the `geak_v3` and `geak_v3_triton` READMEs: docker-first setup/run (sglang image, GEAK v3.2.2 installed in-container, git-worktree parallelism so there's no docker-in-docker), AMD gateway auth (`AMD_LLM_API_KEY` + `GEAK_USER`) plus a personal-key option, and AKA-as-source-of-truth.

## Could you handle these in a follow-up PR?

- **Make `geak_eval` measure/report performance like the other tasks.** Right now every `geak_eval` task comes back `WARN` on the task_validator's performance check, so they aren't directly comparable to the rest of the benchmark. Two things:
  1. **Timing methodology**: the harnesses use `warmup=50 / iters=200` and report the **median** (some use `triton.testing.do_bench`, one uses `5/20`).
     Could you switch them to the recommended **10-warmup / 100-measured mean**?
  2. **JSON performance report**: the harnesses only print latency to stdout (`GEAK_RESULT_LATENCY_MS`). Could they also write a JSON report with per-test-case timings, the way the vllm/rocmbench tasks produce `build/performance_report.json`? That way results are parsed the same as other tasks and `benchmark_method_consistent` is populated correctly.
- **MXFP4 tasks need gfx950.** `fused_mxfp4_quant_moe_sort` and `fused_moe_mxfp4` require gfx950/gfx1250 and fail on gfx942 (MI300/MI308) — and their prompt says "MI300X" while the kernel needs gfx950. Could you decide whether to gate/skip them on gfx942 or run them on gfx950, and reconcile the
  prompt/target arch?
- **Simplify GEAK patch handling.** GEAK v3.2.2 auto-applies and commits the winning patch itself, so the launcher's `_apply_best_patch` is now a redundant fallback and `geak_summary.json` can say `patch_applied: false` even on a
  successful run. Worth simplifying/clarifying.

PR #32, opened 2026-04-16
Multi-shape test cases and torch2hip evaluation improvements

sharareh-y:
### Summary
- **Added multi-shape test cases** to all `torch2hip` gpumode tasks, enabling evaluation across a variety of input dimensions instead of a single fixed shape.
- **Cleared HIP reference files** — `.hip` files are now initialized as empty strings so the agent starts from scratch rather than being seeded with a pre-written kernel.
- **Fixed `hip2hip` correctness checks** — aligned the correctness evaluation with the multi-shape performance test cases to ensure consistency between correctness and performance scoring.
- **Removed stale tasks** (`LearnedPositionalEncoding`, `SoftNLL`) that were no longer part of the benchmark suite.
- **Updated orchestration** — minor adjustments to `main.py`, ` src/performance.py` to support the new evaluation flow.

### Affected areas
- `tasks/torch2hip/gpumode/` — 24 tasks updated (correctness checks, perf scripts, pytorch modules, HIP files)
- `tasks/hip2hip/gpumode/` — correctness checks unified with new multi-shape format
- main.py and src/performance.py

### Test plan
- [x] Run a full `torch2hip` benchmark pass on MI300 and verify multi-shape correctness + performance reports generate correctly
- [x] Confirm `hip2hip` tasks still pass correctness with updated checks


irvineoy (changes requested):
I found one blocking issue in the new multi-shape flow.

In the updated correctness/perf scripts we now create separate module and functional input generators and iterate them together, e.g.:
- `tasks/torch2hip/gpumode/10024_Feedforward/eval_tools/correctness_check.py:194-240`
- `tasks/hip2hip/gpumode/Feedforward/eval_tools/correctness_check.py:194-240`
- `tasks/torch2hip/gpumode/10024_Feedforward/eval_tools/cal_kernel_perf.py:310-352`

The problem is that both `get_inputs()` implementations now yield fresh random tensors:
- `tasks/torch2hip/gpumode/10024_Feedforward/pytorch_code_module/py_10024_Feedforward.py:25-44`
- `tasks/torch2hip/gpumode/10024_Feedforward/pytorch_code_functional/py_10024_Feedforward_func.py:60-79`

So `inputs_modu_gen` and `inputs_func_gen` do *not* produce the same values, even with `torch.manual_seed(0)` set before the loop. A quick repro on this branch already gives different first tensors from `next(module.get_inputs())` vs `next(func.get_inputs())`. That means we are no longer doing like-for-like comparisons: correctness can fail because the two models saw different inputs, and the perf report can mix baseline/optimized timings from different random samples.

This looks systematic across the multi-shape torch2hip tasks, and the same pattern was copied into the hip2hip correctness checks. Please generate each test case once and deep-copy/share it across both paths (or otherwise force the two generators to yield identical tensors per case) before we trust the new evaluation results.

PR #38, opened 2026-05-07
feat(hip2hip): inject generic task contract at framework level

mehdi-saeedi:
## Summary

Follow-up to **#35** (review thread). @irvineoy asked that the four generic task constraints — preserve kernel names/signatures, keep launch/config interface compatible, keep the build/test interface intact, handle shared-memory launch sizing — survive the cleanup of the MI300X-specific prompt that previously lived in `tasks/hip2hip/others/points_in_boxes/config.yaml`. Per the agreed plan, this PR adds them at the **framework level** so every hip2hip task gets the same contract, with no per-task YAML duplication.

## No-touch guarantee

This PR is **strictly additive for `task_type == 'hip2hip'` only**. The following task types are **untouched** by this change — they render the exact same prompt before and after:

- `triton2triton` (incl. `rocmbench/*`, `geak_eval/*`, `repository/rocprim/*`)
- `torch2hip`
- `cuda2hip`
- `instruction2triton`
- `repository`

The injection sites in both `src/prompt_builder.py` and `agents/geak_v3/launch_agent.py` are gated by an `if task_type_name == 'hip2hip':` check; nothing else in either file moves.

Other things this PR does **not** touch:
- No task YAML edits (no `config.yaml` modified).
- No new dependencies.
- No changes to evaluator, scorer, task discovery, runner, or worktree handling.
- No changes to `default_cheatsheet.yaml`, the architecture cheatsheets, or the GPU arch-precheck directive.
- The cheatsheet injection added in #35 is preserved verbatim — the contract is appended alongside it, not in place of it.

## What changes

A new `### Task Contract (Generic)` section is rendered into the agent prompt for every `task_type == 'hip2hip'` task. The contract enumerates:

1. **Preserve kernel function names and signatures** — listed by name from the task's own `target_kernel_functions` field so the agent sees exactly which symbols are part of the contract.
2. **Keep the launch / configuration interface compatible** — grid/block dims, stream usage, host-side launch helpers, Python bindings, `extern "C"` shims.
3. **Output must remain directly compilable and runnable** — the task's existing `compile_command` / `correctness_command` / `performance_command` must succeed against the modified source with no edits to the test runner and no extra build flags.
4. **Handle shared-memory launch sizing correctly** when `__shared__` / dynamic LDS allocations are introduced.

The contract is injected from two places, mirroring the section ordering used by each prompt path:

- `src/prompt_builder.py::prompt_builder` — regular agents. Inserted right after the `### Source Code` section.
- `agents/geak_v3/launch_agent.py` — GEAK-v3 (uses `simple_prompt_builder`, bypasses the regular path). Mirrors the cheatsheet injection added in #35.

## Why framework-level (not per-task `prompt.instructions`)

Across the **36 current hip2hip configs**:
- 24 `gpumode/*` already have a generic role-preamble in `prompt.instructions` (no contract bullets).
- 12 `others/*` have `instructions: null` and `cheatsheet: null`.

Adding the contract per-config would require touching 36 YAMLs and inevitably drift over time. Hosting it in one function keeps the contract in lockstep across tasks, single-source-of-truth, and architecture-neutral by construction.

## Diffstat

\`\`\`text
agents/geak_v3/launch_agent.py | +17 -2
src/prompt_builder.py          | +11 -1
src/prompts/task_type.py       | +66
3 files, +94 / -3
\`\`\`

## Test plan

- [x] All 36 hip2hip configs render with the contract present and the correct kernel name(s) listed.
- [x] Sampled triton2triton configs (5/5) do NOT receive the contract (correctly scoped).
- [x] Regular `src/prompt_builder.py::prompt_builder` path verified.
- [x] GEAK-v3 `agents/geak_v3/launch_agent.py` path verified (logs `incl. hip2hip contract` when fired).
- [ ] End-to-end run on a representative `hip2hip` task confirms the agent receives the contract in `task_prompt.md`.
- [ ] Spot-check that an agent attempting to rename a target kernel function sees the contract section in its initial prompt.
- [ ] Spot-check a triton2triton run pre/post this PR — `task_prompt.md` is byte-identical.

### DIFF
--- a/agents/geak_v3/launch_agent.py
+++ b/agents/geak_v3/launch_agent.py
@@ -160,8 +160,15 @@ def launch_agent(eval_config: dict[str, Any], task_config_dir: str, workspace: s
     # minimal, so the cheatsheet (and the arch-precheck directive) are
     # appended here from the shared prompt_builder helpers, mirroring the
     # section layout used by src/prompt_builder.py::prompt_builder.
+    #
+    # Also inject the hip2hip task contract for hip2hip tasks so the
+    # constraints (preserve names/signatures, launch interface, build
+    # interface, shared-memory sizing) reach GEAK-v3 agents too. The
+    # contract is hosted at the framework level rather than per-task to
+    # keep all hip2hip configs uniform.
     try:
         from src.prompt_builder import _load_cheatsheet, _gpu_arch_precheck_prompt
+        from src.prompts import task_type as _task_type_module
         with open(task_config_dir, "r") as _f:
             _task_config = yaml.safe_load(_f) or {}
         _task_type_name = _task_config.get("task_type", "")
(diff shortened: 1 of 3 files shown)


irvineoy:
Thanks!

PR #39, opened 2026-05-07
feat(flydsl2flydsl): add flydsl2flydsl task type with FlyDSL kernel test examples

johayang-amd:
### Summary

Adds a new `flydsl2flydsl` task type for benchmarking FlyDSL (FlyDSL Python DSL) kernel optimization on AMD GPUs.

**Framework changes:**
- `src/prompt_builder.py` — route `flydsl2flydsl` to the new task type persona
- `src/prompts/task_type.py` — add FlyDSL optimization specialist persona
- `src/prompts/cheatsheet/default_cheatsheet.yaml` — register FlyDSL cheatsheet
- `src/prompts/cheatsheet/flydsl_cheatsheet.md` — FlyDSL optimization guide for agent prompt context

**4 initial kernel tasks** (sourced from [FlyDSL](https://github.com/ROCm/FlyDSL)):
- `rmsnorm_kernel` — RMSNorm with float32 accumulation
- `layernorm_kernel` — LayerNorm with mean/variance reduction
- `softmax_kernel` — Numerically stable softmax using exp2
- `fused_rope_cache_kernel` — Fused Q/K RoPE rotation + KV cache write

Each task includes `kernel.py`, `test_kernel_harness.py` (supports `--correctness` / `--full-benchmark`), and `config.yaml`. Harnesses write `build/performance_report.json` for evaluator compatibility.

### Scope

This PR is strictly additive. All existing task types are unaffected — no existing configs, evaluator, scorer, or agent templates are modified.

### Testing

- [ ] Validate tasks with `task_validator`
- [ ] End-to-end benchmark run


irvineoy:
Two medium follow-ups worth addressing before merge:

1. The `compile_command` for the new FlyDSL tasks currently only does an AST parse of `kernel.py`. That catches syntax errors, but it does not validate that FlyDSL can be imported, that the builder/JIT path is valid, or that the generated module can be constructed. Could we strengthen this to at least import the kernel and instantiate/build a small representative module, if runtime cost is acceptable?

2. The PR introduces a new FlyDSL task family, but I do not see an install/dependency story for FlyDSL in `requirements.txt`, docs, or task setup. If FlyDSL is expected to be preinstalled in the benchmark image, please document that assumption; otherwise please add the required dependency/setup instructions so these tasks are reproducible in a clean environment.

johayang-amd:
> Two medium follow-ups worth addressing before merge:
> 
> 1. The `compile_command` for the new FlyDSL tasks currently only does an AST parse of `kernel.py`. That catches syntax errors, but it does not validate that FlyDSL can be imported, that the builder/JIT path is valid, or that the generated module can be constructed. Could we strengthen this to at least import the kernel and instantiate/build a small representative module, if runtime cost is acceptable?
> 2. The PR introduces a new FlyDSL task family, but I do not see an install/dependency story for FlyDSL in `requirements.txt`, docs, or task setup. If FlyDSL is expected to be preinstalled in the benchmark image, please document that assumption; otherwise please add the required dependency/setup instructions so these tasks are reproducible in a clean environment.

@irvineoy > Hi, thanks for the review and the `sys.exit` fix in `9966369`. I've applied the same pattern to the 3 newly added harnesses as well. Below are the updates:
> 
> **1. `compile_command` strengthened** (`8cb0d1d`)
> 
> Replaced the AST parse with a real module import + build call for all tasks. For example, softmax now runs:
> ```
> python3 -c "from kernel import build_softmax_module; build_softmax_module(2, 256, 'f16'); print('compile ok')"
> ```
> This catches import errors, missing FlyDSL installation, and JIT compilation failures at the compile step. The runtime cost is minimal (sub-second for small shapes).
> 
> **2. FlyDSL dependency documented + difficulty coverage expanded** (`663ae19`)
> 
> Added `tasks/flydsl2flydsl/README.md` documenting the prerequisite (`pip install flydsl`) and a verify command. Also expanded the task set to cover difficulty levels: 3 Easy (softmax, rmsnorm, layernorm), 1 Medium (fused_rope_cache), and 3 Hard (flash_attn_func, hgemm_splitk, pa_decode_fp8). The Hard kernels are sourced from [ROCm/FlyDSL](https://github.com/ROCm/FlyDSL) with all external dependencies inlined into `kernel.py`. Each harness follows the same `sys.exit` pattern from your commit.

irvineoy:
Thank you for your contribution

PR #40, opened 2026-05-11
add CODEOWNERS, CONTRIBUTING.md, SECURITY.md

sinarafati-amd:
## Summary

- Adds `.github/CODEOWNERS` with team + individual contributor handles.
- Adds `CONTRIBUTING.md` at the repo root — project-tailored contribution guide mirroring the team's standard structure (Before You Start / Setup / Workflow / Code Style / Testing / Filing Issues / Security / Suggested Contributions / License).
- Adds `SECURITY.md` at the repo root — vulnerability-disclosure policy pointing to GitHub Private Vulnerability Reporting and the AMD Product Security portal, with a project-specific Scope section.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,12 @@
+# Team slugs: https://github.com/orgs/AMD-AGI/teams
+# Syntax:     https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+# Last matching rule wins. Prefer GitHub Teams over individuals.
+
+# Default — every file requires project-team approval
+# (any of: team OR any listed contributor satisfies the rule)
+*                            @AMD-AGI/brain-tio @irvineoy @sharareh-y @sinarafati-amd @sharonzhou 
+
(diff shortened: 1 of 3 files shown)


irvineoy:
I found a couple of CODEOWNERS issues that should be fixed before relying on this for enforcement:

1. `.github/CODEOWNERS:7,10-12` currently references invalid/unusable team owners.

GitHub CODEOWNERS validation reports:
- `@AMD-AGI/brain-tio` is unknown / not visible / or does not have write access to this repo.
- `@AMD-AGI/brain-tio-leads` is unknown / not visible / or does not have write access to this repo.

This means the intended ownership rules may not be enforced correctly, especially the stricter lead-owner rules.

2. `.github/CODEOWNERS:12` uses `/src/core/`, but this repo does not currently have a `src/core/` directory.

If the goal is to require lead approval for evaluator/scoring/runtime/core framework logic, this rule currently matches nothing. Please update it to cover the actual paths that need stricter ownership, e.g. relevant files/directories under `src/`.

There is also a small trailing whitespace issue on line 7 flagged by `git diff --check`.

PR #44, opened 2026-06-24
Add campaign20 head kernel benchmark tasks (20 cases).

chushic:
Introduce baseline harness tasks under tasks/campaign20/baseline for GEAK head-kernel benchmarking.


irvineoy:
Thanks for adding these tasks! 

Before merging, could you please confirm whether all newly added tasks have been run through the project’s Task Validator Agent, using the latest available Codex or Claude Code, and share the validation results in the PR?

One more question on task organization: these tasks are placed under the new tasks/campaign20/baseline category, while their task_type values are existing categories (triton2triton, hip2hip, torch2hip, etc). I wonder should these tasks live under the existing triton2triton / hip2hip task categories so they follow the current task organization and discovery conventions?

irvineoy:

irvineoy (changes requested):
Thanks for adding these campaign20 benchmark tasks. I did a quick pass and there are a few blockers that should be addressed before we can merge this.

1. Some of the new tasks appear to be broken or not self-contained:
   - `gemm_a8w8_blockscale`, `moe_stage1`, and `moe_stage2` have `source_file_path` entries that do not exist in the task directory, so they will fail source-file validation.
   - Some tasks list external production ops as `target_kernel_functions` rather than functions defined in the provided source files, e.g. `moe_gemm_fp8_blockscale` and `paged_attention_decode`. This means the agent does not really have a local editable kernel target.
   - A few runners/harnesses depend on hardcoded external paths or installed packages, such as `/sgl-workspace/aiter` / `GEAK_REPO_ROOT`, which breaks the self-contained task requirement.

2. `main` has recently gone through a large refactor, so this PR needs to be rebased onto the latest `main` and adjusted to the current task structure.

3. Please run the `task_validator` agent for these tasks and attach the validation reports. Only tasks that pass validation should be added to the codebase.

4. These 20 tasks are currently placed under a new category folder: `tasks/campaign20/baseline`. Based on their configs, 7 are `hip2hip` tasks and 13 are `triton2triton` tasks. Please place them under the existing task categories instead of creating a new top-level category. If you want to run exactly this 20-task subset, you can create your own local config that lists these tasks explicitly.

PR #48, opened 2026-06-25
Docker-first benchmarking with CUDA-graph timing; prune tritonbench/unused agents; harden task_validator

irvineoy:
## Summary

This PR moves AgentKernelArena onto a **Docker-first benchmarking flow** with
**CUDA-graph-based performance timing**, slims the repo by removing the
`tritonbench` task suite and several unused agent integrations, and **hardens
the task_validator** (Codex backend, Python-env handling, prompt fixes, and
support for `repository` tasks). It also fixes a batch of concrete task/env
failures surfaced by a full validation run.

### 1. Docker-first benchmarking
- New `scripts/docker_benchmark.sh`: runs benchmarks inside pinned ROCm/SGLang
  images (`lmsysorg/sglang:*-rocm720-mi30x|mi35x`), with GPU device passthrough,
  non-root user mapping, host-tool mounts (node/codex/claude/cursor), and writable
  scratch caches (`/tmp/*` for torch/triton/matplotlib/MIOpen).
- Arch auto-detection (gfx942/gfx950) and per-arch image overrides.

### 2. CUDA-graph performance timing
- `src/performance.py`, `src/testcases.py`, `src/evaluator.py`,
  `src/module_registration.py` reworked for graph-capture timing.
- Each task's `scripts/task_runner.py` (≈200 tasks) gains a shared
  `_benchmark_cuda_graph_or_events()` helper: CUDA-graph capture with warmup +
  measured replays, automatic fallback to CUDA-event timing when capture is
  unsupported, and per-case timing reports.

### 3. Repository slimming
- Removed the entire `instruction2triton/tritonbench` suite (**184 tasks**).
- Removed 6 unused agent integrations: `SWE_agent`, `geak_hip`,
  `geak_optimagentv2`, `geak_ourllm_kernel2kernel`, `openevolve`,
  `single_llm_call`. Maintained agents: `claude_code`, `codex`, `cursor`,
  `task_validator`.

### 4. task_validator hardening
- Codex backend support, correct Python-environment propagation, and prompt
  fixes (`agents/task_validator/{launch_agent.py,validation_prompt.py,agent_config.yaml}`).
- Added support for `task_type: repository`: enum entry, optional source/target
  hints, recursive source lookup for cloned repos, lenient self-contained check
  for upstream build deps, and `build/*.json` + `baseline_perf.yaml` result-template
  mapping.

### 5. Bug fixes from a full validation run
- **Env** (`scripts/docker_benchmark.sh`): writable MIOpen cache (fixes
  `miopenStatusUnknownError` lockfile failures); bind-mount host `/usr/bin/time`
  (fixes exit-127 compiles); `/opt/venv/bin` ahead on `PATH` so bare `python3`
  has torch (fixes repository `task_runner` imports).
- **GELU** kernel computed ReLU → replaced with exact erf-GELU.
- **InnerProd** / **MaskedLanguageModel** kernels: double-precision reduction to
  match torch BLAS over large C/H (were failing `rtol=1e-4/atol=1e-5`).
- **ball_query**: `run_correctness` now compares against the existing
  `cpu_reference` golden (the prior within-radius check passed an all-zeros kernel).

### 6. Docs
- Updated `README.md`, `CONTRIBUTING.md`, `docs/how-to/add-task.md`,
  `docs/reference/api-reference.md` for the docker flow and `repository` task type.

## Testing
- Validator + scripts syntax-checked; kernel and HIP edits reviewed against the
  reference implementations.
- Full GPU re-validation of the touched tasks


sharareh-y (commented):
Thanks for the Docker-first direction and the concrete kernel/validator fixes. I went through the diff in detail — overall this is a solid refactor, but I have several concerns I'd like addressed before merge.

---

### Docker requires all three agent CLIs

`build_docker_args()` hard-requires Claude Code, Cursor Agent, and Codex paths/auth dirs, even if `config.yaml` only uses one agent (e.g. Codex). That means `make docker-shell` / `make docker-run` fail for anyone who hasn't installed and logged into all three.

Could we make agent mounts conditional on the configured `agent.template` (or an env flag), and keep strict checks only for `check-agents`?

---

### CUDA-graph timing scope is overstated in the PR description

The summary says ~200 `task_runner.py` files get `_benchmark_cuda_graph_or_events()`, but the actual coverage is narrower:

- **`triton2triton/vllm/*/scripts/task_runner.py`** (~118 files) — CUDA graph helper added
- **`*/rocmbench/*/performance_utils_pytest.py`** (62 files) — same helper inlined
- **`hip2hip/gpumode/*`** — unchanged; still uses `eval_tools/cal_kernel_perf.py` (10 warmup / 100 CUDA events)
- **`hip2hip/others/ball_query`** — correctness fix only; timing unchanged

Timing methodology is now inconsistent across task categories. Please update the PR description to reflect the real scope, or extend graph timing to hip2hip `cal_kernel_perf.py` if uniform methodology is the goal.

---

### Validator performance check conflicts with the new timing approach

`validation_prompt.py` still requires **PASS** only when warmup=10, measured=100, and an average is verified. CUDA-graph tasks use adaptive `n_repeat`, median of graph replays, and emit `benchmark_method: cuda_graph` — so Check 6 will systematically **WARN** even on successful runs.

Please update the validator prompt to accept the CUDA-graph methodology (with its documented warmup + median replay semantics) as PASS, or explicitly document that perf methodology WARN is expected for graph-timed tasks.

---

### Duplicated benchmark helper (~120 copies)

`_benchmark_cuda_graph_or_events()` and `_measure_cuda_event_fallback()` are copy-pasted into each task file (~100+ lines each), with no shared module under `src/`.

Suggest extracting to something like `src/benchmark_cuda_graph.py` and importing from task runners / `performance_utils_pytest.py`, so future timing fixes don't require editing ~180 files.

---

### Stale documentation

README and `install.md` are updated for Docker-first, but several docs still treat `python main.py` / `make setup` as the primary path:

- `docs/README.md` (install table)
- `docs/how-to/run-evaluation.md` (resume examples)
- `docs/how-to/agents.md`, `docs/examples/examples.md`
- `docs/reference/api-reference.md`

Please align these with the Docker-first workflow, or add a clear note that Docker is the canonical path.

---

### Legacy venv path weakened

The `Makefile` no longer seeds `torch` into auto-generated `requirements.txt`, and docs de-emphasize `make setup`. Contributors following the legacy venv workflow can easily end up without a ROCm PyTorch build.

If we still support the venv workflow, please keep `torch` in the generated requirements (or document the exact manual install steps in one place). If Docker is the only supported path going forward, say that explicitly and trim or relocate the legacy Makefile targets.

sharareh-y:
## Risk: baseline and optimized can be timed with different methods, fabricating speedups

The new CUDA-graph timing in `performance_utils_pytest.py::_measure_times` silently falls back to per-call CUDA-event timing on any capture exception:

```python
except Exception as exc:
    ...
    metadata.update({
        "benchmark_method": "cuda_event_fallback",
        "benchmark_fallback_reason": f"cuda_graph_failed: ...",
    })
    return _measure_cuda_events(callable_fn, config.repetition), metadata
```

Because the baseline (`measure_baseline`, original kernel) and the optimized kernel (`measure_performance`, agent output) are measured in **separate invocations**, nothing guarantees they use the *same* method. If one side captures into a CUDA graph (host-overhead-free) while the other falls back to event timing (which includes per-launch + sync overhead), the reported `speedup = baseline_mean / optimized_mean` reflects the **measurement-method delta, not kernel quality**. For small kernels the launch-overhead floor is large, so this can manufacture multi-x "speedups" (or hide regressions) for a functionally identical kernel.

This is a correctness issue in the metric itself, separate from environmental noise.

### Suggested fix

Make the timing method consistent per baseline/optimized pair:

- Record `benchmark_method` for both sides and, if they differ, either (a) re-measure both with the common fallback method, or (b) mark the comparison invalid and fail loudly instead of emitting a speedup.
- Surface `benchmark_method` for baseline and optimized in the final report so mixed-method comparisons are visible.

irvineoy:
Validated all existing tasks and fixed all the issues I found.

sharareh-y:
Thanks for the changes, one thing I noticed looking at the perf helpers — the copy/sync setup feels pretty easy to break and I wanted to flag it.

Right now `tools/perf/` is the source of truth, but tasks don't import it — `sync_perf_helpers.py` copies it out into every task instead.
A few ways this can silently go wrong:

- If someone edits a task copy directly, it just quietly diverges from `tools/perf` — only `--check` in CI would catch it, assuming that's wired up as a required gate (is it?).
- If someone edits `tools/perf` but forgets to run `make sync-perf-helpers`, you end up with ~250 stale copies and no warning.
- The vLLM path worries me more — it doesn't even copy a file, it string-injects the helper functions between the `# >>> AKA-GENERATED` markers inside each `task_runner.py`. If a marker gets removed or edited, the sync silently skips (or errors on) that file and it drifts on its own.

Basically the "single source of truth" only holds as long as everyone remembers to run the sync and never hand-edits a copy — and when it breaks, it breaks silently at runtime (you just get wrong speedups instead of an error). Not a blocker, just wanted to put it on your radar.

sharareh-y:
One more small thing, related to the perf-helper sync setup: the workflow is documented, but only in `tools/perf/README.md`, which is basically orphaned — nothing in `CONTRIBUTING.md`, `docs/how-to/add-task.md`, the root README, or `docs/reference/benchmark-methodology.md` links to it or mentions it. So a contributor adding a new task today has no way to discover that the timing helpers are centrally generated and that you have to run the sync script.

Right now the only safety net for that is:
- the inline `# >>> AKA-GENERATED ... run make sync-perf-helpers >>>` marker comments inside the ~180 task copies (only seen if you happen to open a copy), and
- `make check-perf-helpers` / `--check` in CI (assuming it's wired up as a required gate).

A couple of cheap doc fixes that would close the gap:

1. Add a short note to `docs/how-to/add-task.md` (and a line in `CONTRIBUTING.md`) along the lines of: *"The performance-timing block in `task_runner.py` / `performance_utils_pytest.py` is generated from `tools/perf/`. Don't hand-edit it — edit the canonical file there and run `make sync-perf-helpers`. Run `make check-perf-helpers` before pushing."*
2. Link `tools/perf/README.md` from `docs/reference/benchmark-methodology.md` (it already links the other direction, so this just makes the round-trip discoverable).

Not a blocker — just want contributors to actually find the sync workflow instead of relying on them stumbling into the marker comments.

irvineoy:
Nice catch. I re-implemented those parts. It should be clearer now. @sharareh-y

.github/workflows/perf-helpers.yml
```diff
@@ -0,0 +1,15 @@
+name: perf-helpers
```
sinarafati-amd, line 1: tasks/instruction2triton/tritonbench is completely removed right? shall we let other teams know just in case there is any dependency.

src/scripts/docker_benchmark.sh
```diff
@@ -0,0 +1,645 @@
...
+
+    docker_args=(run --rm --entrypoint bash)
+    unset _MOUNTED_TARGETS
+    declare -gA _MOUNTED_TARGETS=()
+    if [[ "$interactive" == "1" && -t 0 ]]; then
+        docker_args+=(-it)
+    fi
+
+    docker_args+=(
+        --ipc=host
```
sinarafati-amd, line 341: The container runs with --ipc=host --network=host --privileged --cap-add=SYS_ADMIN --cap-add=SYS_PTRACE --security-opt=seccomp=unconfined while executing model-generated code. This is host-level access for an autonomous agent and might be risky. maybe lets remove privileged flag

src/scripts/docker_benchmark.sh
```diff
@@ -0,0 +1,645 @@
...
+    # non-root user (so it cannot apt-install it). Bind-mount the host binary
+    # read-only so commands that invoke `/usr/bin/time` do not fail with 127.
+    if [[ -x /usr/bin/time ]]; then
+        add_mount /usr/bin/time /usr/bin/time ro
+    fi
+
+    if [[ -e "$HOST_HOME/.gitconfig" ]]; then
+        add_mount "$HOST_HOME/.gitconfig" "$HOST_HOME/.gitconfig" ro
+    fi
+    if [[ -d "$HOST_HOME/.ssh" ]]; then
```
sinarafati-amd: lets not mount .ssh its risky

.github/workflows/perf-helpers.yml
irvineoy: Yes, you are right. I will let other teams know about it. And. I also released a version before this PR, so they should be able to find the old code if they need it.

src/scripts/docker_benchmark.sh
irvineoy: Nice catch

src/scripts/docker_benchmark.sh
irvineoy: I think I encountered some wired issue when I did not provide enough permission to the agent. I will double check it later.

PR #69, opened 2026-07-21
feat(forge): support Forge agent and add image_kernel task type

jiaqiang-dot-liu:


tasks/image_kernel/aiter_gemm/scripts/task_runner.py
```diff
@@ -0,0 +1,250 @@
...
+
+    return torch.matmul(case["x"], case["w"].T)
+
+
+CASES = [
+    dict(m=256, n=512, k=512, dtype_str="bfloat16"),
+    dict(m=128, n=256, k=512, dtype_str="float16"),
+]
+
+PERF_CASES = [
```
irvineoy, line 185: **[P1] Validate the scored performance shapes for correctness**

The new `image_kernel` runners use separate `CASES` and `PERF_CASES`, but `run_correctness()` only validates `CASES`. Here, the small correctness shapes differ completely from the 4096/8192 shapes that are actually scored. A kernel can therefore remain correct for the validation shapes while producing incorrect results—or specializing invalid behavior—for the scored shapes and still receive a valid performance score. The same pattern appears in all seven new image-kernel runners.

Please validate every `PERF_CASES` configuration against its reference at least once before timing it, or include the performance configurations in the correctness suite.

agents/forge/drivers/arena_task_adapter.py
```diff
@@ -0,0 +1,114 @@
...
+    """Reuse Arena's correctness evaluation; emit the forge allclose contract."""
+    try:
+        from src.evaluator import evaluate_correctness
+    except Exception as e:  # noqa: BLE001
+        print("allclose: False")
+        print(f"error: cannot import Arena evaluator (arena_root={arena_root!r}): {e}")
+        return 0
+
+    task_config_data = _load_task_config(task_config)
+    passed, err = evaluate_correctness(Path(workspace), task_config_data)
```
irvineoy, line 88: **[P1] Recompile each Forge candidate before correctness and benchmarking**

The generic adapter calls `evaluate_correctness()` and `measure_performance()` directly, without first running `evaluate_compilation()`. This evaluates stale binaries for tasks where compilation is a separate step. For example, the existing `hip2hip/others/matrix_multiplication` and `mla_decode` runners build an executable in `compile` mode, while their correctness and performance modes only execute the existing binary. Arena compiles the baseline before launching Forge, but source edits made during a Forge iteration do not rebuild that binary.

Consequently, Forge may validate and benchmark the pre-agent binary rather than the edited kernel. Please make each candidate follow a compile → correctness → benchmark sequence, or otherwise guarantee that build artifacts are regenerated after every source change.

agents/forge/launch_agent.py
```diff
@@ -0,0 +1,676 @@
...
+
+
+def _resolve_gpu_arch(eval_config: dict[str, Any]) -> str:
+    """Map the Arena ``target_gpu_model`` to a gfx arch (defaults to gfx942).
+
+    The arch is derived solely from the run's ``target_gpu_model`` (always set by
+    the API / config); no environment override is needed.
+    """
+    model = str(eval_config.get("target_gpu_model", "")).upper()
+    return _GPU_ARCH_MAP.get(model, "gfx942")
```
irvineoy: **[P1] Do not silently map unsupported GPU models to `gfx942`**

`_resolve_gpu_arch()` defaults every unrecognized GPU model to `gfx942`. RDNA4 is supported elsewhere in this repository as `gfx1201`, but it is absent from this mapping, so a Forge run targeting RDNA4 would be given the MI300 architecture. This can produce invalid build flags, misleading agent guidance, or kernels optimized for the wrong ISA.

Please reuse the repository's shared GPU architecture resolution, use the architecture already detected by `setup_rocm_env`, or fail explicitly for unknown models instead of silently falling back to `gfx942`.

agents/forge/launch_agent.py
```diff
@@ -0,0 +1,676 @@
...
+        threading.Thread(target=read_stream, args=(process.stderr, stderr_lines, "[FORGE STDERR]", logger.warning), daemon=True),
+    ]
+    for t in threads:
+        t.start()
+
+    try:
+        process.wait(timeout=timeout_seconds)
+    except subprocess.TimeoutExpired:
+        logger.warning(f"Forge loop timed out after {timeout_seconds}s; terminating")
+        process.terminate()
```
irvineoy: **[P2] Terminate the complete Forge process tree on timeout**

The Forge command is launched with `shell=True`, and this timeout handler terminates only the shell process. The `kernel-agents` process and its Claude/GPU subprocesses may survive after the shell exits, continuing to consume GPU resources or modify the workspace while Arena runs `git checkout` and performs final evaluation.

Please launch the command as an argument list with `shell=False`, create a separate process group/session, and terminate the entire group on timeout.

src/jit_rebuild.py
```diff
@@ -0,0 +1,220 @@
...
+    if not strs:
+        return
+    # Only C/C++ HIP kernels have the prebuilt-.so shadowing problem; forcing a
+    # rebuild for a Triton (.py) task would just recompile aiter's C++ for nothing.
+    if not any(s.endswith(_CPP_EXTS) for s in strs):
+        return
+    joined = " ".join(strs)
+
+    if "aiter" in joined:
+        os.environ.setdefault("AITER_REBUILD", "1")
```
irvineoy: **[P2] Scope `AITER_REBUILD` to the current task**

This modifies the parent process environment and never restores it. Workers execute multiple tasks in the same process, so an AITER HIP task can leave `AITER_REBUILD=1` enabled for subsequent AITER Triton tasks, triggering unnecessary C++ rebuilds and potentially causing large runtime regressions. In addition, `setdefault()` does not actually force rebuilding if the inherited environment already contains `AITER_REBUILD=0`.

Please scope this variable to the relevant task subprocesses, or save and restore its previous value when the task finishes. When rebuilding is required, set it explicitly to `1`.

irvineoy (changes requested):
Overall, the direction of this PR looks valuable, especially the Forge integration, image-backed kernel tasks, harness integrity checks, and centralized timing. However, I found several issues that can cause stale or incorrect kernels to be evaluated. I recommend addressing the P1 findings before merging.

tasks/image_kernel/aiter_gemm/scripts/task_runner.py
jiaqiang-dot-liu: Fixed. run_correctness() now validates PERF_CASES in addition to CASES. It builds the combined list [*CASES, *scored_cfgs] (where scored_cfgs are the shapes extracted from PERF_CASES) and runs each through the aiter-vs-torch reference check, so every scored shape is verified against its reference before it can earn a performance score. This is applied consistently across all seven image-kernel runners, not just this one.

agents/forge/drivers/arena_task_adapter.py
jiaqiang-dot-liu: Fixed. Both do_correctness() and do_bench() now call _rebuild_edited_source() first, which invokes Arena's evaluate_compilation() to regenerate build artifacts from the current (agent-edited) source before validating/timing. A failed rebuild aborts the step (correctness reports allclose: False, bench returns a non-zero exit) so a stale pre-edit binary is never validated or benchmarked. JIT-only tasks (no compile_command) skip the step harmlessly. This gives each candidate a compile → correctness → benchmark sequence.

agents/forge/launch_agent.py
jiaqiang-dot-liu: Fixed. _resolve_gpu_arch() no longer contains a hardcoded map defaulting to gfx942. It now reuses Arena's shared resolution — the same priority as src.preprocessing.setup_rocm_env: first the real hardware arch from rocminfo (_detect_gfx_arch_from_rocminfo()), then the configured target_gpu_model looked up in the single source of truth (the architecture map in default_cheatsheet.yaml, which covers RDNA4→gfx1201, MI300/MI325/MI355X, ...). An unknown model now raises ValueError with a clear message instead of silently falling back to gfx942.

agents/forge/launch_agent.py
jiaqiang-dot-liu: Fixed. The command is now launched from the argv list with shell=False and start_new_session=True, so the forge-loop, kernel-agents, and the Claude/GPU subprocesses share one process group led by the child PID. On timeout, _terminate_process_group() signals the whole group (os.killpg with SIGTERM, then SIGKILL after a grace period) rather than just the leader, so no descendants survive to hold the GPU or edit the workspace while Arena runs git checkout and final scoring. (The shlex.quote-joined cmd string is kept only for human-readable logging.)

src/jit_rebuild.py
jiaqiang-dot-liu: Fixed. AITER_REBUILD is no longer written to os.environ. apply_jit_rebuild() returns the override as a dict ({"AITER_REBUILD": "1"}), and callers in evaluator.py / performance.py apply it only to the specific build subprocess via run_command(..., extra_env=rebuild_env). Since it never touches the parent env, it cannot leak into a later task run by the same worker. It's also set explicitly to "1" (not setdefault) so an inherited AITER_REBUILD=0 is still correctly overridden when a rebuild is required.

PR #72, opened 2026-07-24
Add GEAK v4 kernel workflow agent integration

irvineoy:
## Summary

- add a registered `geak_v4` agent that runs GEAK's deterministic `kernel_workflow.js` through a persistent Claude Agent SDK client
- add Docker setup, read-only GEAK checkout mounting, persistent SDK dependencies, preflight checks, MI300 quickstart config, and user documentation
- run GEAK on a disposable task snapshot and import only a Director-accepted patch for the one declared source file
- handle background Workflow lifecycle, disk artifact recovery, physical-to-logical GPU mapping, bounded output, and process-group cleanup

## V1 task contract

The initial adapter supports `hip2hip`, `triton2triton`, and `flydsl2flydsl` tasks with exactly one ordinary source file. It rejects protected test/harness/config paths, symlinks, hard links, multiple sources, binary patches, path traversal, and create/delete/rename/mode or multi-file changes.

The adapter was developed against GEAK commit `4965d5b2ccde927925c8c5501a25c1233daa52eb` (`v4.0.0-102-g4965d5b`).

## Integrity boundary

- GEAK receives a disposable workspace copy and `apply_to_original=false`
- the scored workspace is fully manifested before and after Workflow execution
- Director, Workflow, and imported patch provenance must all match the pinned `eval/final_patch.diff`
- result/audit JSON uses symlink-safe atomic writes, artifact directories are inode-pinned, and normal runs mount the persistent Python dependency target read-only
- this is a fail-closed integrity boundary, not an OS sandbox; that limitation is called out in the agent README

## Validation

- `python -m pytest -q tests` — **84 passed**
- `bash tests/test_docker_benchmark.sh` — passed
- Python compile checks, shell syntax checks, YAML parsing, `git diff --check`, and Make target dry runs — passed
- handoff mapping against the real GEAK checkout and pinned `kernel_workflow.js` — passed
- cached gfx950 container setup/import smoke for `claude-agent-sdk==0.2.127`, including `ClaudeSDKClient`, `ClaudeAgentOptions`, and `cli_path` — passed without contacting Claude
- two independent read-only reviews found no remaining blocker after patch-provenance, lifecycle, dependency-mount, outer-Git-worktree, and symlink-write fixes

## Not tested

A live Claude/GEAK optimization run was intentionally not executed because this machine has no paid Claude Code account. The included one-task quickstart is the remaining authenticated end-to-end check.

## Suggested review focus

- compatibility of the handoff and normalized result schema with the pinned GEAK revision
- Claude SDK background-task lifecycle and terminal artifact semantics
- whether GEAK maintainers want to widen the V1 single-source task contract in a follow-up

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 
 SHELL := /bin/bash
 
-.PHONY: help docker-shell docker-check-agents docker-smoke docker-run docker-parallel-run docker-setup-flydsl \
+.PHONY: help docker-shell docker-check-agents docker-smoke docker-run docker-parallel-run docker-setup-flydsl docker-setup-geak \
         check-docker-runner check-evaluator check-held-out check-visualization \
         visualization-build visualization-serve visualization-run \
         sync-perf-helpers check-perf-helpers materialize-perf-workspace \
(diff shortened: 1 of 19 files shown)


irvineoy:
Merged the #73

PR #73, opened 2026-07-27
Add GEAK v4 agent integration

yueliu14:
<html>
<body>
<html><head></head><body><h2>Summary</h2><p>Integrates the GEAK v4 agent as a kernel-optimization backend in AgentKernelArena, targeting AMD MI300 (<code inline="">gfx942</code>).</p><p>GEAK runs through an SDK bridge that drives a multi-role workflow pipeline:</p><p><code inline="">author → profile → engineer(s) → verify → integrator → director validation</code></p><p>AgentArena then independently re-measures every result using its centralized evaluator, ensuring an authoritative, apples-to-apples performance score.</p><h2>What's Included</h2><ul><li><p><code inline="">agents/geak_v4/</code> SDK bridge:</p><ul><li><p><code inline="">workflow_runner.py</code></p></li><li><p><code inline="">launch_agent.py</code></p></li></ul></li><li><p>Integration of GEAK's <code inline="">kernel_workflow.js</code> pipeline with the Arena agent interface.</p></li><li><p>Example configuration:</p><ul><li><p><code inline="">quickstart_geak_v4_mi300.yaml</code></p></li></ul></li></ul><h2>Test Results</h2><p><strong>Environment:</strong> AMD MI300<br><strong>Test scope:</strong> 9-task smoke run across 6 GPUs</p><p>The nine tasks consist of:</p><ul><li><p>5 <code inline="">hip2hip/others</code> tasks</p></li><li><p>4 <code inline="">triton2triton/geak_eval</code> tasks</p></li></ul><p>The results below are based on <strong>Arena's authoritative independent re-measurement</strong>, rather than GEAK's internal director-reported numbers.</p>

| Category | Result
-- | --|
| hip2hip/others (5) | 5/5 sped up, average speedup 8.97×: ball_query 18.78×, roiaware_pool3d 17.91×, assign_score_withk 5.58×, points_in_boxes 1.30×, matrix_multiplication 1.29×|
| triton2triton/geak_eval (4) | 3/4 sped up, average speedup 1.55×: llama_ff 2.50×, topk 1.50×, refk_identity 1.34×, fused_qkv_rope 0.86× (regression, correctly flagged)|

</body></html>
</body>
</html>

</body></html>
</body>
</html>


irvineoy (commented):
I found two integration issues that should be addressed before relying on the new GEAK v4 quickstart.

example_configs/quickstart_geak_v4_mi300.yaml
```diff
@@ -0,0 +1,11 @@
+# Minimal first-run example for GEAK v4 on MI300/MI300X (gfx942).
+# Complete agents/geak_v4/README.md setup before running it.
+agent:
+  template: geak_v4
```
irvineoy, line 5: **[P1] Wire this template into the Docker-first runner before exposing the quickstart.**

For `make docker-run CONFIG=example_configs/quickstart_geak_v4_mi300.yaml`, `resolve_required_agents` returns the literal `geak_v4`, while `mount_agent` only provisions `codex`, `claude_code`, and `cursor`. The container therefore gets neither the Claude binary nor its auth state. The runner also does not forward/mount the host's `GEAK_V4_WORKFLOW_DIR`, and no supported setup step installs `claude-agent-sdk` in the runtime Python. Following this quickstart through the project's supported Docker path will consequently fail during workflow/Claude/SDK preflight. Please map `geak_v4` to Claude Code provisioning/checks and add an explicit GEAK checkout mount/env plus SDK setup or preflight.

agents/geak_v4/workflow_runner.py
```diff
@@ -0,0 +1,876 @@
...
+    director_valid = _valid_director_validation(validation, eval_dir)
+    primary_metric_valid = speedup is not None
+    if (
+        accepted
+        and gained
+        and patch_exists
+        and director_valid
+        and not workflow_contract_invalid
+    ):
+        status = "ok"
```
irvineoy, line 676: **[P2] Do not report success unless the accepted patch reached the Arena workspace.**

This `ok` condition does not require `validation["applied_to_original"] == "true"`. With otherwise valid `accepted`/`pass` artifacts, a 1.2x speedup, a non-empty pinned patch, and `applied_to_original="false"`, `normalize_result()` still returns `status="ok"`; the runner exits zero and the launcher only logs a warning. Arena then re-scores the unchanged baseline, silently discarding a validated optimization. Please treat that contract mismatch as an error (or have the launcher reliably import the patch) and cover it with a regression test.

yueliu14: Fixed in a75d9a1. normalize_result() now takes require_applied (set when the run requested apply_to_original=="true"). With it, an accepted+gained result is only "ok" when applied_to_original=="true"; otherwise status="error" with a reason that Arena would re-score the unmodified baseline.

example_configs/quickstart_geak_v4_mi300.yaml
yueliu14: Fixed in a75d9a1. resolve_required_agents/normalize_check_agents now map geak_v4→claude_code (container gets the Claude binary + auth). The runner mounts+forwards the host's GEAK_V4_WORKFLOW_DIR, and make 
 docker-setup-geak (container_setup_geak) installs claude-agent-sdk into the persistent pip user-base via pip install --target + PYTHONPATH (venv rejects --user; /opt/venv unwritable by the non-root user), 
short-circuiting when the image already ships it; the run flow runs setup+check preflight for geak_v4. Also forwards the host Claude gateway auth (ANTHROPIC_AUTH_TOKEN+ANTHROPIC_BASE_URL+model/TLS vars). Verified end-to-end on MI300X: docker-check-agents → authMethod=oauth_token, and a real docker-run reaches GEAK's optimize/profile loop.

PR #74, opened 2026-08-03
feat(forge): pass task identity and workload metadata to forge-loop

jiaqiang-dot-liu:
Forward stable kernel identity and workload metadata to Forge while keeping scoring independent from external publication state.

### DIFF
--- a/agents/forge/launch_agent.py
+++ b/agents/forge/launch_agent.py
@@ -21,13 +21,15 @@ kernel with the task's own compile/correctness/performance commands.
 
 from __future__ import annotations
 
+import hashlib
+import json
 import logging
 import os
+import re
 import shlex
 import shutil
 import signal
 import subprocess
-import sys
 import threading
 import time
 from pathlib import Path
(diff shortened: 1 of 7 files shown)


agents/forge/launch_agent.py
```diff
@@ -533,6 +851,215 @@ def _render_driver_shim(drivers_dir: str, workspace: str, task_config: str, aren
...
+    ]
+    if source_files:
+        cmd.extend(["--source-files", ",".join(str(path) for path in source_files)])
+    if target_functions:
+        cmd.extend(["--target-functions", ",".join(str(name) for name in target_functions)])
+    if logical_operator:
+        cmd.extend(["--operator-name", logical_operator])
+    if framework:
+        cmd.extend(["--framework", framework])
+    if shapes:
```
irvineoy, line 916: [P1] The `shape-v2` workload identity built here is no longer consumed by the current KernelForge CLI. On current `main`, `--shapes-json` is a hidden legacy/checkpoint-only option and `--workload-key` is explicitly ignored ([declarations](https://github.com/AMD-AGI/KernelForge/blob/fc2e8a87d9db3a045df3b2c10128996621389c0a/src/kernel_agents/cli.py#L730-L731), [warnings/behavior](https://github.com/AMD-AGI/KernelForge/blob/fc2e8a87d9db3a045df3b2c10128996621389c0a/src/kernel_agents/cli.py#L884-L895)). Consequently, two tasks with the same source/operator but different shape regimes will not be isolated by the key promised in this PR's README. Please migrate this handoff to the current campaign/invocation-spec identity contract, or pin/probe a compatible KernelForge version and fail clearly instead of silently degrading.

PR #77, opened 2026-08-03
Use graph replay across all kernel benchmarks

### LINKED ISSUE #53: Timing issue in flydsl tasks
The `flydsl2flydsl` performance harness
(`tasks/flydsl2flydsl/*/test_kernel_harness.py`) times each iteration like this:
```python
start.record()
launch_fn(x, output, M)
end.record()
torch.cuda.synchronize()
times.append(start.elapsed_time(end))
```

launch_fn does a lot of host-side dispatch work between the two events. At that point the GPU stream has nothing else to run, so the GPU sits idle waiting for the CPU to enqueue the kernel, and that idle gap falls right between start and end — so elapsed_time(start, end) also counts the host launch overhead. For small ops / small shapes this host overhead dominates the measured time (the actual GPU compute is often just a few microseconds).

Why this is a problem
Does not reflect real deployment: real inference runs these kernels under CUDA/HIP Graph (graph-mode serving), which eliminates the per-launch host overhead via graph replay. So for small shapes the current metric does not represent how the op actually behaves in production.
Dilutes real GPU optimization: with a fixed host-launch floor, an agent's genuine GPU-compute improvements look negligible against that launch overhead.
Incentivizes the wrong thing: an agent can lower the measured time by optimizing the Python/launch path (e.g. caching the compiled dispatch state, reusing packed argument pointers) instead of the kernel itself — a "speedup" that disappears entirely once CUDA Graph is enabled.

Suggested fix
Measure only the actual GPU compute time, so the agent optimizes the kernel rather than the launch path.

irvineoy:
## Summary

- Make CUDA/HIP Graph replay the default timing path across all 426 task configurations: 220 canonical Python benchmarks, 6 native HIP graph drivers, 139 vLLM/image adapters, and 61 ROCmBench adapters.
- Centralize adaptive replay batching, graph priming, state preparation, method metadata, and observable timed-output replay in the protected benchmark helpers.
- Treat each baseline case's method as immutable. A graph-capable baseline cannot be downgraded to Event timing by a candidate capture failure; method mismatches receive no performance score.
- Restore mutable hip2hip inputs before every reference and candidate invocation, using independent buffers so neither side consumes state written by the other.
- Exclude GEAK method mismatches from geometric means instead of treating them as 1.0x, and propagate an explicit aggregate consistency marker to scoring.
- Validate the exact graph executable that was timed. Native matrix and MLA drivers now use deterministic nonzero inputs, poison outputs, replay, and compare results before reporting performance.
- Align timed workload boundaries: FlyDSL top-k paths use the same output-allocation contract, ROI scratch is caller-owned, and LoRA/output resets run through `prepare_fn` outside timing.
- Rebase the PR on current `main` (including #82) and preserve its timed-output protections in the canonical helper.

## Task-validator update

- Bring the validator up to the current 426-task repository contract, including repository/image tasks, per-task command timeouts, materialized helpers, legacy generation/starters, exact workspace path resolution, and protected co-located harnesses.
- Expand validation from 10 to 12 checks with dedicated `benchmark_integrity` and `harness_integrity` reviews. Benchmark validation now rejects host/CPU timing, partial or anonymous case sets, missing/mixed methods, candidate-triggered fallback, unrestored state, asymmetric timed work, captured reset/scratch/allocation, zero-only inputs, and unchecked Graph replay output.
- Add schema-v2 framework finalization: command attempts and exit codes are authoritative, TIMEOUT is a failure, missing/malformed checks fail closed, and `overall_status` is recomputed instead of trusted from model output.
- Track backend return code/timeouts, atomically finalize reports with a digest marker, rerun partial/legacy reports on resume, aggregate prior reports during resume, and return nonzero from the final CLI gate when validation fails.

## CI and validation

- `python -m pytest -q tests`: 228 passed, 2 GPU-only tests skipped.
- `make check-perf-helpers`: all 426 configurations classified and helpers in sync.
- `python -m compileall -q src tasks tests`: passed.
- `git diff --check`: passed.
- Added `benchmark-integrity`, which runs the helper audit, complete CPU/mock suite, and Python compilation on pull requests.
- Added an isolated `benchmark-gpu-smoke` workflow using the pinned ROCm image. It covers stateful preparation, exact timed replay, forced-Event rejection, and native matrix/MLA replay validation, with logs uploaded as artifacts when an isolated `rocm`/`gfx950` runner is enabled.
- Real MI355X smoke: 2/2 Python Graph tests passed; native matrix and MLA drivers compiled and reported `cuda_graph` after nonzero replay-output validation.

## Compatibility notes

- Validator reports from the old unversioned schema are intentionally not treated as resumable completion; rerun validation to generate schema-v2 reports.
- Historical benchmark results without explicit method-consistency metadata should remain version-isolated; reprocessing them under this scoring policy intentionally removes their performance component.
- Materialized helper tasks should be run through AgentKernelArena. Direct task-source execution still requires helper materialization.
- Graph-first timings are a new benchmark version and should not be compared directly with older host-launch/Event results.

Closes #53


src/testcases.py
```diff
@@ -530,6 +615,95 @@ def calculate_average_speedup(
...
+
+    A ``mixed:...`` value is an aggregate set, not a stable case-to-method
+    mapping.  Equal mixed strings therefore cannot prove that the same shapes
+    used the same timer in the baseline and optimized runs.  Refuse those
+    aggregate comparisons instead of silently scoring a potentially swapped
+    graph/event assignment.
+    """
+
+    mismatches: List[Dict[str, Any]] = []
+    comparable_methods = {'cuda_graph', 'cuda_event_fallback'}
```
sharareh-y, line 645: **Blocking: this makes 14 of the 21 `image_kernel` tasks unscoreable.**

Requiring `benchmark_method ∈ {cuda_graph, cuda_event_fallback}` on both sides is right, but most `image_kernel` runners emit `benchmark_method` only nested under a `"metadata"` key, and` _build_metadata_from_case` never flattens it — so the parser sees `None` on both sides and every pair becomes `missing_or_unknown_benchmark_method`.

Reproduced against this branch's own `src`, using the exact row shape from `tasks/image_kernel/mi355x_vllm_triton_unified_attention/scripts/task_runner.py:399-404`:


```
consistent = False
mismatches = [{"reason": "missing_or_unknown_benchmark_method", ...} x2]
score      = 120.0     # a real 2x speedup; should be 320
```
An AST scan of all 21 runners gives 14 nested-only (all `mi355x_vllm_*` except `aiter_mxfp4` and `kda_linear`, plus both `mi355x_sglang_*`) and 7 flat. Two of the flat ones already carry the comment "_Flat, not nested: src/testcases.py reads `benchmark_method` from the top level_" — so the trap was known; this PR converted 7 runners without applying it to the other 14.

src/harness_guard.py
```diff
@@ -76,14 +86,97 @@ def _sha256(path: Path) -> str:
...
+    except (OSError, UnicodeDecodeError, SyntaxError):
+        return "invalid-python:" + _sha256(path)
+
+    for node in tree.body:
+        if (
+            isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))
+            and node.name in editable_targets
+        ):
+            node.body = [ast.Pass()]
+    canonical = ast.dump(tree, annotate_fields=True, include_attributes=False)
```
sharareh-y, line 227: **Blocking: this rejects normal kernel optimization on 30 ROCmBench tasks.**

 These tasks keep the kernel and the pytest harness in one file, so the file is now protected with only the target function's *body* left editable. Everything else in the file is frozen. Measured on the real tasks:

 | What the agent does | Result | Tasks affected |
 |---|---|---|
 | Edits the kernel body | works | — |
 | Adds an `import` | **run rejected** | all 30 |
 | Adds a `@triton.jit` helper kernel | **run rejected** | all 30 |
 | Tunes `@triton.autotune` configs | **run rejected** | 6 of 30 |

 All three are things we want agents to do. Tuning the autotune configs especially — on `hard/gemm` the tile sizes `BLOCK_SIZE_M/N/K` are `constexpr` parameters that *only* come from that config list, so freezing it removes tile-size selection, which is most of the available speedup on that task.

 The cause is that `_sha256_python_harness` masks `node.body` only. A decorator sits in `node.decorator_list`, and imports and new functions are separate top-level nodes, so all of them stay in the hash.

 The consequence is harsh: `verify_workspace_harness` raises, `run_task` swallows it, and no `task_result.yaml` is written at all. The task then disappears from the aggregate report entirely (`_collect_all_tasks_from_run` filters on that file), and the run exits nonzero. An agent that did good work loses everything with no partial credit and no diagnostic in the report.


 Suggested fix: invert the rule. Instead of "hash the whole module, unmask one function body," hash the whole module *except* the nodes that are clearly kernel implementation:

 - the declared `target_kernel_functions`, as **whole nodes** (decorators, signature, and body)
 - any top-level `@triton.jit` function the agent adds
 - top-level `import` / `from` statements

 Everything else, the `test_*` functions, their `parametrize` decorators, module-level shape constants, and any bare top-level statement, stays hashed. That still blocks the attacks you're targeting, while letting the three edits above through.

sharareh-y:
Hi, thanks for the updates! Blocker 1 is fully fixed. Blocker 2 is fixed except for one case, so there's one thing left to fix. Breakdown below.

**Blocker 1 (nested `benchmark_method`) — fixed.** `_build_metadata_from_case` now promotes nested `benchmark_*` keys with row-level values winning. The unified-attention row shape that previously scored 120.0 on a real 2× speedup now scores 320.0, and all 14 nested-only `image_kernel` runners are covered by that one parser change. `test_nested_benchmark_metadata_is_promoted_for_scoring` and `test_flat_benchmark_metadata_wins_over_nested_compatibility_value` lock in both the promotion and the precedence rule. Nothing further needed.

**Blocker 2 (harness guard scope) — mostly fixed, one gap left.** Imports and `@triton.jit` helpers now work; I re-probed 8 ROCmBench tasks and both edits are accepted on all of them. `.task-venv` / `.venv` / `__pycache__` exclusion is a good catch on top.

The autotune half only reaches 1 of the 6 affected tasks, though. Masking the target node recovers config tuning only when the config list is written inline in the decorator. Five tasks call a helper instead, and that helper is an ordinary top-level `FunctionDef` — not the target, not `@triton.jit` — so it stays in the digest:

| Task | Target decorator | Edit configs |
|---|---|---|
| `hard/gemm` | `@triton.autotune(configs=[triton.Config(...), ...])` | **ACCEPTED** |
| `medium/layernorm` | `@triton.autotune(configs=get_autotune_config(), ...)` | **REJECTED** |
| `medium/rmsnorm_fwd` | `@triton.autotune(configs=get_autotune_config(), ...)` | **REJECTED** |
| `medium/softmax` | `@triton.autotune(configs=get_autotune_config(), ...)` | **REJECTED** |
| `hard/multreduce_matmul_dot_kernel` | `@triton.autotune(configs=get_triton_dot_autotune_configs(), key=get_triton_autotune_key())` + `@triton.heuristics(get_triton_heuristics())` | **REJECTED** |
| `hard/triton_multreduce_matmul_kernel` | `@triton.autotune(configs=get_triton_multreduce_autotune_configs(), ...)` + `@triton.heuristics(...)` | **REJECTED** |

So the tasks where tile-size and warp-count selection matters most still can't be tuned — on those five, `BLOCK_SIZE_*`, `num_warps`, `num_stages`, and `waves_per_eu` all come out of that one helper.

Could we extend the mask by one hop in `src/harness_guard.py:190-199`? When a target node is dropped, walk its `decorator_list` for called names and drop those top-level `FunctionDef`s too. That covers `get_autotune_config`, `get_triton_dot_autotune_configs`, `get_triton_multreduce_autotune_configs`, `get_triton_autotune_key`, and `get_triton_heuristics` without a task-specific allowlist. The hop is bounded to functions a declared target's decorator actually calls, and those exist solely to configure the kernel.

Worth noting `test_colocated_kernel_imports_structure_and_triton_helpers_are_editable` puts the config list inline, which is the `hard/gemm` shape that already worked — a case using `configs=get_autotune_config()` would have caught this.

PR #79, opened 2026-08-04
Enhance Forge integration and add MI355X image-kernel tasks

jiaqiang-dot-liu:


tasks/image_kernel/mi355x_vllm_triton_paged_attention_2d/scripts/task_runner.py
```diff
@@ -0,0 +1,288 @@
...
+    sys.modules[EDIT_MODULE_NAME] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+def _make(case: dict, correctness: bool = False) -> dict:
+    torch = _torch()
+    params = dict(case["params"])
+    num_seqs = min(params["num_seqs"], 8) if correctness else params["num_seqs"]
+    ctx_len = min(params["ctx_len"], 256) if correctness else params["ctx_len"]
```
irvineoy: **[P1] Validate the shapes that are actually scored**

Correctness clamps every case to `num_seqs=8` and `ctx_len=256`, while performance scores `num_seqs=64` at context lengths 1024, 2048, and 3072. With `block_size=16`, the gate covers 16 pages per sequence but the scored runs cover 64, 128, and 192 pages. An optimization that breaks long-context traversal or the 64-sequence launch can therefore pass correctness and still be scored. Please validate the scored shapes, or add equivalent coverage for these size-dependent paths.

tasks/image_kernel/mi355x_vllm_ck_a8w8_blockscale_gemm/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+  kernel_kind: ck
+  source_owner: aiter
+  workload:
+    source: session_cases.json
+    primary_case: deepseek-flash-k003
+compile_command:
+- python3 scripts/task_runner.py compile
+correctness_command:
+- python3 scripts/task_runner.py correctness
+performance_command:
```
irvineoy, line 20: **[P1] Provide a real `--profile-run` contract for the CK tasks**

This task and the two sibling CK tasks (`mi355x_vllm_ck_cktile_moe_2stage` and `mi355x_vllm_ck_moe_2stage`) do not ship `scripts/forge_driver.py`, so Forge generates the generic adapter. That adapter uses `parse_known_args`, silently ignores `--profile-run`, and executes full correctness. Current KernelForge profile preflight treats a zero exit as a verified contract, so task preparation is skipped and profiling includes compilation, correctness, and reference work instead of a kernel-only path. Please ship profile-capable drivers for all three tasks, or make the generic adapter explicitly reject/implement `--profile-run`.

example_configs/pr69_image_kernel_validator_mi355x.yaml
```diff
@@ -7,8 +7,6 @@ tasks:
   - image_kernel/aiter_mha_batch_prefill
   - image_kernel/aiter_pa_decode
   - image_kernel/aiter_pa_ragged
-  - image_kernel/aiter_rmsnorm
```
irvineoy, line 10: **[P2] Keep the validator task names in sync with the renames**

Removing these two deleted tasks is not sufficient: this PR also renames all five remaining `aiter_*` task directories, while the config keeps their old paths. Running `_discover_tasks()` with this list now returns an empty dictionary, so the example silently validates zero tasks. Please update or replace the entries; note that the renamed directories are `mi300x_*`, so mechanically substituting those names into this MI355X config would also be misleading.

tasks/image_kernel/mi355x_vllm_tilelang_mhc_fused_post_pre/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+image_repo_path: /usr/local/lib/python3.12/dist-packages/vllm/model_executor/kernels/mhc
+repo_subdir: mhc
+repository_language: tilelang
+source_file_path:
+- tilelang.py
+target_kernel_functions:
+- mhc_fused_post_pre_tilelang
+kernel_identity:
+  logical_operator: mhc_fused_post_pre
+  kernel_kind: tilelang
```
irvineoy, line 11: **[P2] Avoid silently routing TileLang to the Triton fellow**

The Forge backend mapping and valid-backend set do not include `tilelang`, and the current KernelForge fellow registry has no TileLang backend. Consequently this task falls through to `triton-fellow`, giving it the wrong specialist prompt and KB backend identity. Please add a matching TileLang fellow/backend before enabling Forge for this task, or fail fast instead of silently selecting Triton.

tasks/image_kernel/mi355x_vllm_triton_paged_attention_2d/scripts/task_runner.py
jiaqiang-dot-liu: Fixed in 4941ed2. Correctness now validates the exact scored shapes (num_seqs=64, ctx_len=1024/2048/3072), covering 64/128/192 pages per sequence. The reduced 8×256 shape is retained only for the compile smoke test. The standalone driver and regression coverage were updated as well.

tasks/image_kernel/mi355x_vllm_ck_a8w8_blockscale_gemm/config.yaml
jiaqiang-dot-liu: Fixed in 2b96600. All three CK tasks now ship profile-capable forge_driver.py implementations. --profile-run prepares one full performance case and runs only the target operator path, without invoking correctness, references, or benchmarking. Regression tests verify the profile behavior for all three tasks.

example_configs/pr69_image_kernel_validator_mi355x.yaml
jiaqiang-dot-liu: Fixed in 3c7c8d5. The MI355X validator now references all current mi355x_* image-kernel tasks instead of the renamed legacy entries. Task discovery also fails explicitly when a configured selector matches no task, preventing zero-task validation runs from succeeding silently.

tasks/image_kernel/mi355x_vllm_tilelang_mhc_fused_post_pre/config.yaml
jiaqiang-dot-liu: Fixed in 45739f4. Arena no longer maintains a KernelForge backend allowlist or falls back to Triton for unknown backends. It now forwards the backend declared by the task configuration directly, so kernel_kind: tilelang becomes tilelang-fellow. KernelForge is responsible for deciding whether that fellow is supported.

PR #82, opened 2026-08-09
Make image-kernel correctness check what performance actually scores, and add three tasks

jiaqiang-dot-liu:
Two things: close a scoring hole in the existing image-kernel tasks, and add three new tasks that do not have it.

## 1. Fix the existing tasks

A kernel could collect a speedup on these tasks without the scored code path ever being checked. Two independent gaps allowed it.

**The scored shapes were never validated.** Correctness built inputs at truncated shapes while performance used the full session shapes, so on any operator that branches on size the two modes exercised different code. `fused_moe_gptq_awq` validated at 16 tokens / 8 experts / top-2 and scored at 7211 tokens / 384 experts / top-8; `unified_attention` validated at ctx_len 128 and scored at 2048. A kernel that skipped work on the long-prompt path alone passed correctness and kept the speedup.

**The scored invocation was never observed.** Correctness and performance are separate calls, and nothing ever looked at what the timed call produced: the benchmark owns the captured graph and returns only a duration. A kernel could tell the two apart and do less work in the one that is scored.

What changed:

- The correctness/performance shape switch is removed from the ten scored operators. `_make` / `_prepare` no longer take a `correctness` flag, so the two modes cannot diverge by construction. Compile keeps a cheap smoke shape through an explicit helper, which is what `paged_attention_2d` already did.
- `vllm_cuda_graph_block` gains an optional `timed_run` collector that binds a handle to the buffers the timed unit last wrote, plus a way to re-run that same unit. Every return path binds it, including the four fallbacks. Performance then asserts on those buffers after perturbing the inputs and poisoning the outputs, so work the scored path skips cannot hide behind a correctness call that took a different branch.
- Four references walked (token, slot) pairs in Python and synchronised on the routing tensor every step, which is what made a full-shape check impractical. They now batch per expert or per query chunk, so `sparse_attn` at 7211 queries and `fused_moe_gptq_awq` at 7211 tokens / 384 experts both validate in seconds.
- `fused_moe_gptq_awq` rotates the whole case suite three times instead of running each case once. A single pass samples a kernel once against a fixed seed, so an implementation that is only intermittently wrong passes whenever the defect does not fire.
- Two tasks opt out of the timed-output assertion and say why in place: `kda` advances state in place and a captured graph chains several invocations, so its golden has to carry state across repeats; `mxfp8_grouped_gemm` freezes GEMM2's activation at setup, so the timed result cannot witness whether GEMM1 ran.
- Three runners copied from the multi-operator template drop the branches their operator cannot reach.

One unrelated fix rides along because it blocked the E2E validation: aiter treats any non-zero `AITER_REBUILD` as "rebuild this module", never checking whether the source changed, and the profiler re-spawns the driver once per counter pass. Forcing the rebuild only once an editable source no longer matches its in-image original takes a single `--profile-run` from 19s to 4s, and the profiling stage of an arena + forge-loop run from not completing after 540s to finishing in 210s.

## 2. Add three tasks

`paged_attention_rocm`, the Gemma4 fused MoE and the Gemma4 unified attention, brought over from `feat/add-new-image-kernel-0804` with the shape split already closed rather than inherited. All three truncated correctness inputs the same way: the MoE shrank every dimension (7218 tokens / 128 experts / top-8 / hidden 2816 down to 16 / 8 / 2 / 512), and both attention tasks cut num_seqs 64 to 8 and ctx_len up to 2048 down to 256.

`unified_attention_gemma4` is the sharpest case and not only a "less work" hole. Its reference clips the window with `lo = max(0, ctx_len - window)`, so at the scored ctx_len 2048 with sliding_window 1024 it drops half the context, while a correctness ctx_len of 256 makes `lo` zero and clips nothing. The sliding-window behaviour was therefore never validated at all, and a kernel that got it wrong could still score. `paged_attention_decode` has a milder version: ctx_len also sets the split-K partition count, so a shortened context changes which reduction path runs.

## Validation

On gfx950, compile / correctness / performance pass at full shapes for the nine vllm tasks this container can build, and arena + forge-loop runs clean through baseline, task-prepare skip, source map, profiling and KB warm start on each. Timings are unchanged from before the change. The remaining four (`mxfp8_linear`, `mxfp8_grouped_gemm`, `mxfp4_moe_2stage`, `kda_linear_attn`) need images this container cannot provide.

The gates now catch things they used to admit. Four stored KB solutions for `mhc_fused_post_pre`, claiming 11x to 3808x, fail full-shape correctness; the best surviving entry is 1.667x. The top KB solution for `fused_moe_gptq_awq` folds the moe_align prefix sum into the histogram kernel behind a cross-workgroup software barrier with no sound ordering guarantee, and caches the counter buffer in a module global that the barrier is supposed to re-zero; it produced wrong results 5 times in 12 full-shape runs. Suite rotation takes detection on it from 3/8 to 6/8 while the pristine kernel stays at 8/8, so the gate gains samples without gaining false rejections, at no measurable wall-clock cost.

### DIFF
--- a/src/tools/perf/vllm_cuda_graph_block.py
+++ b/src/tools/perf/vllm_cuda_graph_block.py
@@ -42,8 +42,9 @@ class _TimedRun:
 
     Under CUDA-graph timing the buffers are captured once and every replay
     writes to those same addresses, so ``outputs`` keeps tracking replays. Under
-    event-timing fallback each call allocates fresh buffers, so ``outputs`` is
-    only populated once ``rerun`` has run.
+    event-timing fallback the measured outputs cannot be observed reliably, so a
+    benchmark that requests this collector fails closed instead of validating a
+    separate post-timing invocation.
     """
 
     def __init__(self):
@@ -82,19 +83,12 @@ def _benchmark_cuda_graph_or_events(
 ):
     import torch
 
-    def _bind_direct_call():
-        # Fallback paths allocate fresh outputs per call, so there is nothing to
-        # alias; the caller gets them by re-running the same unit once.
-        if timed_run is None:
-            return
-
-        def _call_once():
-            out = fn()
-            if torch.cuda.is_available():
-                torch.cuda.synchronize()
-            return out
-
-        timed_run._bind(_call_once)
+    def _reject_unobservable_fallback(reason):
+        if timed_run is not None:
+            raise RuntimeError(
+                f"{reason}; timed_run requires an observable CUDA-graph replay "
+                "and cannot validate a separate post-timing invocation"
+            )
 
     # A captured graph whose measured per-iteration time falls below this floor is
     # treated as empty (it recorded no device work) and rejected in favour of
@@ -116,23 +110,23 @@ def _benchmark_cuda_graph_or_events(
     }
 
     if not torch.cuda.is_available():
+        _reject_unobservable_fallback("CUDA is unavailable")
         times = _measure_cuda_event_fallback(fn, repetition)
         metadata.update({
             "benchmark_method": "cpu_timer_fallback",
             "benchmark_effective_repeats": int(repetition),
             "benchmark_fallback_reason": fallback_reason or "cuda_unavailable",
         })
-        _bind_direct_call()
         return sum(times) / len(times), metadata
 
     if not use_cuda_graph:
+        _reject_unobservable_fallback("CUDA-graph timing is disabled")
         times = _measure_cuda_event_fallback(fn, repetition)
         metadata.update({
             "benchmark_method": "cuda_event_fallback",
             "benchmark_effective_repeats": int(repetition),
             "benchmark_fallback_reason": fallback_reason or "cuda_graph_disabled",
         })
-        _bind_direct_call()
         return sum(times) / len(times), metadata
 
     try:
@@ -178,13 +172,13 @@ def _benchmark_cuda_graph_or_events(
 
         graph_mean = sum(retry_times) / len(retry_times)
         if graph_mean < empty_graph_floor_ms:
+            _reject_unobservable_fallback("CUDA graph captured no device work")
             times = _measure_cuda_event_fallback(fn, repetition)
             metadata.update({
                 "benchmark_method": "cuda_event_fallback",
                 "benchmark_effective_repeats": int(repetition),
                 "benchmark_fallback_reason": fallback_reason or "empty_cuda_graph_capture",
             })
-            _bind_direct_call()
             return sum(times) / len(times), metadata
 
         metadata.update({
@@ -214,6 +208,11 @@ def _benchmark_cuda_graph_or_events(
             torch.cuda.synchronize()
         except Exception:
             pass
+        if timed_run is not None:
+            raise RuntimeError(
+                "CUDA-graph capture failed; timed_run cannot validate the "
+                "separate CUDA-event fallback invocation"
+            ) from exc
         for _ in range(min(3, max(1, int(warmup)))):
             fn()
         torch.cuda.synchronize()
@@ -223,5 +222,4 @@ def _benchmark_cuda_graph_or_events(
             "benchmark_effective_repeats": int(repetition),
             "benchmark_fallback_reason": f"cuda_graph_failed: {type(exc).__name__}: {str(exc)[:160]}",
         })
-        _bind_direct_call()
         return sum(times) / len(times), metadata
--- a/src/tools/perf/vllm_cuda_graph_block.py
+++ b/src/tools/perf/vllm_cuda_graph_block.py
@@ -31,6 +31,43 @@ def _measure_cuda_event_fallback(fn, repetition):
     return times_ms
 
 
+class _TimedRun:
+    """Handle on the exact invocation a benchmark measured.
+
+    Timing and correctness are otherwise separate invocations, so a kernel can
+    tell them apart and do less work in the one that is scored. Passing this
+    collector to the benchmark makes the scored invocation itself observable:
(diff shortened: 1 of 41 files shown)


irvineoy (commented):
A few inline comments on the remaining scoring-integrity and edit-scope gaps.

src/tools/perf/vllm_cuda_graph_block.py
```diff
@@ -41,9 +78,24 @@ def _benchmark_cuda_graph_or_events(
...
+        if timed_run is None:
+            return
+
+        def _call_once():
+            out = fn()
+            if torch.cuda.is_available():
+                torch.cuda.synchronize()
+            return out
+
+        timed_run._bind(_call_once)
```
irvineoy: This fallback reintroduces the separation this PR is trying to close. `_measure_cuda_event_fallback` discards every result, and `_bind_direct_call` validates an additional invocation after timing has finished. An implementation can force graph capture to fail and use host/module state to behave differently on that later call, so the event-timed calls themselves remain unobserved. Could we fail closed whenever `timed_run` is requested, or otherwise retain and validate an output produced inside the timed fallback path?

tasks/image_kernel/mi355x_sglang_triton_mxfp8_grouped_gemm/scripts/task_runner.py
```diff
@@ -275,10 +325,16 @@ def run_correctness() -> None:
 
 
 def run_performance() -> None:
+    # Sibling tasks additionally validate the timed invocation itself. That is
```
irvineoy: This leaves the original scored-vs-validated gap open for this task. `_run_gemms` discards GEMM1’s output, GEMM2 consumes the setup-time `act_q`, and the performance path validates neither result. A conditional/stateful implementation can therefore skip or corrupt GEMM1 in the scored call while the separate `_fused_output` correctness call still passes. Could `_run_gemms` surface both outputs so the timed replay can poison and independently check each, without adding activation/quantization to the measured region?

tasks/image_kernel/mi355x_vllm_triton_fused_moe_gemma4/config.yaml
```diff
@@ -0,0 +1,46 @@
...
+    the int4 fused_moe_kernel_gptq_awq that the sibling task
+    mi355x_vllm_triton_fused_moe_gptq_awq targets. Gemma4 lands here because the ROCm
+    AITER unquantized MoE backend rejects its GELU_TANH activation, so the oracle
+    falls back to Triton. The workload is the session''s per-rank geometry under TP=2
+    with EP=1: 128 experts, top-8, hidden 2816, intermediate 352, activation
+    gelu_tanh, BF16, at 64 (decode), 1080 and 7218 (chunked-prefill) tokens. One
+    harness call covers both per-layer launches, the w13 gate_up GEMM and the w2 down
+    GEMM. Worth knowing: the session ran with vLLM''s fallback MoE config because
+    configs/E=128,N=352,device_name=AMD_Instinct_MI355_OAM.json does not exist, and
+    the harness reproduces that; configs/ is part of the editable tree, so adding a
```
irvineoy: This conflicts with Forge’s edit scope. Forge builds its complete allowlist from `source_file_path + editable_sources`, but this task declares only `fused_moe.py`; the generated program will tell the agent to edit only that file while this prompt says `configs/` is editable. The proposed config file also does not exist yet. Please either add support/allowlisting for creating the tuned config under `configs/`, or remove this optimization path from the task instructions.

src/tools/perf/vllm_cuda_graph_block.py
jiaqiang-dot-liu: Fixed in bdfc1e4.

When `timed_run` is requested, the benchmark now fails closed if CUDA is unavailable, graph timing is disabled, the graph captures no device work, or graph capture fails. The CUDA-event fallback remains available only to callers that do not require observation of the timed invocation.

This removes the post-timing direct call, so we no longer validate a different invocation from the one that produced the score. I also added regression coverage for the fail-closed paths.

tasks/image_kernel/mi355x_sglang_triton_mxfp8_grouped_gemm/scripts/task_runner.py
jiaqiang-dot-liu: Fixed in bdfc1e4.

`_run_gemms` now returns both GEMM1 and GEMM2 outputs. The CUDA graph retains aliases to both timed outputs; the validation path poisons both buffers, replays the captured graph, and compares each output independently against a dequantized reference.

The measured region is unchanged: it still contains only the two grouped GEMM launches. Activation and quantization remain outside timing. The check is wired into the Arena task runner, Forge bench driver, and standalone driver. If graph capture fails, this path now fails closed rather than validating a separate event-timed invocation.

The current container does not include the SGLang runtime, so I could not run the GPU E2E here. Syntax checks and the focused integrity regression tests pass.

tasks/image_kernel/mi355x_vllm_triton_fused_moe_gemma4/config.yaml
jiaqiang-dot-liu: Fixed in bdfc1e4.

I removed the incorrect claim that `configs/` is part of this task's editable scope and that creating the missing tuned config is a supported optimization path. The prompt now only retains the factual note that the original session used vLLM's fallback MoE configuration.

I did not add the missing config to `editable_sources`: the file does not exist, while the current Forge/KernelForge allowlist resolves existing files only and would reject that entry in strict mode. Supporting creation of allowlisted files would require a separate new-file contract rather than overloading `editable_sources`.

PR #94, opened 2026-09-04
Add the rewrite_by_flydsl task type and a SIKL suite covering the GLM-5.2 workload schema

jiaqiang-dot-liu:
## Summary

Adds a task type for reimplementing a framework operator in FlyDSL and scoring the result against that operator's own production implementation, the bridge agents that drive it through KernelForge, and a task suite covering every operator the GLM-5.2 workload schema declares.

**`rewrite_by_flydsl` task type.** A task names the baseline implementation to reimplement, the file the port lands in, and the operator identity. It ships a dual-path measurement driver that KernelForge embeds in the port prompt as the definition of the builder/launch interface, times the baseline under `--ref-bench-mode` and the candidate under `--bench-mode`, and checks correctness in the default mode. The port target is the only agent-editable file.

**`forge` and `forge_rewrite` agents.** Bridges from Arena to KernelForge's `forge-loop` and `forge-rewrite-by-flydsl` subprocesses, with their shared GPU identity, git workspace, subprocess streaming and result parsing extracted into `agents/forge/common.py`. Edits outside a task's declared editable set are named in the scored output rather than discarding the campaign.

**SIKL suite: 21 tasks.** Seventeen a16w16 GEMMs and four MXFP4 routed-expert MoEs, one per schema definition, each scored over the thirteen workload cases its schema declares. The builder receives the case shape and is called once per case, so a port specializes per bucket the way the tuned baseline does; the evaluator refuses a partial case match, so a port that only works at some shapes scores no performance at all.

## Design properties worth review

- **Nothing measured is stored in a task.** The correctness gate is derived per run from how far the production implementation itself sits from the fp32 reference, scaled by a policy constant and floored at the loosest reduction strategy that implementation uses. A recorded tolerance would go stale against a different GPU or framework build, and Arena measures its own baseline every run regardless.
- **One measurement implementation per task.** The harness produces the score and the driver produces the number KernelForge keeps a candidate on. Sharing constants is not enough, so the loops themselves are shared in `scripts/task_measure.py` and neither caller times anything on its own.
- **Candidate independence is checked, not stated.** Importing the framework under test is rejected everywhere; the GEMM tasks additionally reject torch's matrix product, which is literally the baseline at their larger M cases.
- **Post-processing fills the schema's solution slot** from Arena's own evaluation rather than from KernelForge's port verdict, because the launcher returns before `evaluate_kernel`.

## Test plan

- [x] `python3 -m pytest tests/ --ignore=tests/gpu` — 751 passed
- [x] `make check-perf-helpers`
- [x] Every commit passes the suite independently
- [x] On MI355X/gfx950: compile, correctness and benchmark for the extreme GEMM shapes (n32 k6144, n6144 k16384, n28672 k512) and all four MoE tasks; all thirteen cases per task time under CUDA-graph replay
- [x] End-to-end `forge_rewrite` run: KernelForge reports `driver contract OK`, then `PORT OK (attempt 1, SNR=49.3)` with the ported FlyDSL GEMM at 0.93x of the tuned baseline across all thirteen cases


irvineoy:
Implemented the review follow-ups in commit fa90e006:

- Restored fail-closed edit-scope enforcement. If Forge changes a tracked file outside source_file_path or editable_sources, Arena now refuses to score the result. Undeclared untracked scratch is still removed before scoring, and removal failure is fatal.
- Hardened candidate independence checks across all 21 SIKL tasks. Candidates can no longer import task_baseline or other protected task modules to reach the production implementation indirectly. GEMM tasks also reject imported torch product aliases and saved references such as product = torch.matmul.
- Added regression coverage for tracked and renamed out-of-scope files, indirect task-module imports, and both torch alias forms.
- Opened the companion SIKL schema change adding flydsl to SupportedLanguages: https://github.com/AMD-AGI/SIKL/pull/17

Validation:
- 426 passed, 21 skipped in the focused host-side suite. The skipped cases require the real KernelForge package.
- make check-perf-helpers passed.
- Python syntax checks and git diff --check passed.

This host has no compatible ROCm GPU, so task_validator GPU validation was not run here and should remain a pre-merge gate.

irvineoy:
Could you clarify your intended long-term policy for Forge edits outside the task declared source_file_path/editable_sources allowlist?

The PR description still says those edits should be named in the output and scoring should continue, motivated by not discarding a long campaign after it finishes. The current head at fa90e006 instead restores fail-closed behavior and refuses to score when a tracked file outside the allowlist changed.

Which outcome do you want?

1. Fail closed: reject the score whenever an undeclared tracked file changed.
2. Continue scoring: retain the score but mark it as relying on undeclared edits.

If you prefer option 2, how should Arena prevent changes to dispatch, dependencies, harness logic, or the measured implementation from making the score untrustworthy? Would keeping fail-closed enforcement while improving preflight or earlier feedback address the original concern?

Please confirm the intended policy so the implementation, tests, and PR description can be made consistent before merge.

irvineoy:
I agree with the overall direction: Arena should support reimplementing an existing production operator in FlyDSL, while keeping correctness and performance scoring under Arena's control. Before merging, though, I suggest simplifying the task contract and avoiding a schema that is tightly coupled to KernelForge's current rewrite CLI.

### 1. Rename the task type to `operator2flydsl`

I suggest:

```yaml
task_type: operator2flydsl
```

This follows the existing `<source>2<target>` naming convention, clearly identifies FlyDSL as the target, and does not constrain the source implementation to Torch, Triton, HIP, or a particular framework dispatch path. It also keeps the task type independent of KernelForge's internal “rewrite” and “port” terminology.

### 2. Add only one machine-readable task field

Most fields in the current `rewrite:` block either duplicate existing Arena fields or describe how one particular agent runs its search. I suggest reducing the task configuration to:

```yaml
task_type: operator2flydsl

# Complete candidate edit allowlist.
source_file_path:
  - kernel.py

# Symbols the candidate must provide.
target_kernel_functions:
  - build_gemm_a16w16_nt_n6144_k6144_module

# The only new task-level field. It must be task-relative.
rewrite_source_file: sources/tuned_gemm.py

kernel_identity:
  logical_operator: gemm_a16w16_nt_n6144_k6144
  source_owner: aiter

compile_command:
  - python3 test_kernel_harness.py --compile
correctness_command:
  - python3 test_kernel_harness.py --correctness
performance_command:
  - python3 test_kernel_harness.py --full-benchmark

platform_support:
  required_arch: gfx950
  status: active

prompt:
  instructions: >-
    Read sources/tuned_gemm.py. Its production entry point is gemm_a16w16.
    Reimplement the operator in FlyDSL in kernel.py.
```

The intended mapping would be:

- remove `rewrite.port_target`; use `source_file_path[0]`;
- move `rewrite.logical_operator` to `kernel_identity.logical_operator`;
- move `rewrite.source_owner` to `kernel_identity.source_owner`;
- replace `rewrite.port_source` with `rewrite_source_file`;
- infer the source language from the file suffix;
- keep the optional source entry point in the task instructions or driver documentation;
- let the task harness/workload files own the complete case suite rather than placing shapes in rewrite-pipeline configuration.

The existing `kernel_identity` handling would need to be shared by the two Forge integrations and documented as a common contract. Today `forge_rewrite` does not read it, while `rewrite.source_owner` is declared by these tasks but is not consumed.

`rewrite_source_file` should resolve inside the task workspace. Paths such as:

```yaml
port_source: /sgl-workspace/aiter/aiter/tuned_gemm.py
```

bind the task to one image layout and conflict with the isolated-task path contract. If the production source comes from the runtime image, it should be materialized into the task workspace through a declared setup mechanism, after which the task should refer to it by a relative path.

### 3. Keep agent search policy out of the task schema

`snr_threshold` and `max_port_attempts` describe how KernelForge runs a campaign, not what the operator computes or how Arena scores it. They should live in the agent/run configuration.

If the PORT filter must vary with numerical characteristics, the agent can derive a sufficiently loose filter from the task's correctness policy. A different agent executing the same `operator2flydsl` task may have no notion of a “port attempt” or a PORT-stage SNR threshold.

### 4. Do not parse execution metadata from the prompt

The source entry point, reading guidance, and implementation context can be documented in `prompt.instructions` or the driver docstring. The launcher should not parse natural-language prompt text to recover paths or execution parameters.

Machine execution should depend only on structured fields such as:

```text
rewrite_source_file
source_file_path
target_kernel_functions
kernel_identity.logical_operator
```

### 5. Rename the agent to `forge_operator2flydsl`

I also suggest renaming the new agent from `forge_rewrite` to:

```yaml
agent: forge_operator2flydsl
```

This names both the provider and the capability, aligns with the proposed task type, and leaves room for other agents to implement `operator2flydsl`. `forge_rewrite` does not identify the target language and makes KernelForge's current pipeline name look like an Arena-wide capability.

### 6. Migration and validation

If this direction is accepted, the same change should update registration/dynamic loading, prompt and cheatsheet routing, shared Forge identity parsing, task-validator schema checks, public task-authoring/API documentation, and focused regression coverage. The new tasks still need finalized GPU `task_validator` reports with `overall_status: PASS` on compatible hardware.

This restructuring should not change the main functionality of the PR. It keeps the task contract backend-neutral and lets `forge_operator2flydsl` adapt that contract to KernelForge's CLI, rather than embedding KernelForge-specific search controls in every task.

jiaqiang-dot-liu:
Thanks — the restructuring is in, and all six points are addressed. Summary of where each landed, then two things found while validating on hardware that you should see before merge.

### 1. `task_type: operator2flydsl` — done (`ad97421`)

All 21 tasks carry it. The name carrying its own target deleted both special cases it needed: `_EXPLICIT_BACKEND` in the Forge backend map and `EXPLICIT_TARGET_LANGUAGE` in the cheatsheet map. Neither had another entry.

### 2. One machine-readable field — done (`ad97421`, `3e4db59`, `ea99d94`)

The `rewrite:` block is gone. `rewrite_source_file` is literally the only key these tasks add beyond the isolated-kernel schema; `port_target` was a second name for `source_file_path[0]`, and `logical_operator` / `source_owner` moved into `kernel_identity`.

`source_owner` is no longer declared-but-unread. `3e4db59` resolves it through `common.py`'s `_resolve_framework` — the same path the `forge` agent uses, which is the shared contract you asked for — and forwards it as `--framework`. That is not tidying: KernelForge otherwise infers the owning framework from the source file's path, and by the time it looks, the source is a copy in the scratch workspace, so `detect_framework` falls through to `unknown`. Framework is one of the six dimensions of a KB recipe's canonical id, so every recipe these tasks published was being filed under an owner nothing looks for.

`rewrite_source_file` is task-relative in all 21 tasks (`aiter_source/aiter/{tuned_gemm,fused_moe}.py`). This needed no new mechanism: Arena already seeds an in-image tree through `image_repo_path`, which 24 `image_kernel` tasks use and `setup_workspace` reads unconditionally rather than off a task type. `repo_subdir` is `aiter_source`, not `aiter`, because a directory of that name at the workspace root would sit on `sys.path` ahead of the real package and shadow the very implementation these tasks are scored against. The task validator now fails an absolute `rewrite_source_file`, and a test pins both that rule and the no-shadowing one.

### 3. Search policy out of the task schema — done (`ad97421`)

`snr_threshold` and `max_port_attempts` live in `agents/forge_operator2flydsl/agent_config.yaml`. Neither key appears in any of the 21 task configs.

### 4. No execution metadata parsed from the prompt — done (`3e4db59`)

The launcher reads only `source_file_path`, `rewrite_source_file` and `kernel_identity`. `rewrite_source_entry` is deleted rather than relocated: KernelForge calls the source host entry a hint for the port agent and says in `ingest.py` that an unresolved one does not block the pipeline, because the driver owns how the reference and the baseline are invoked. The entry now travels only as prose — the prompt sends the agent to `scripts/forge_driver.py`, whose docstring names it — and a test pins both halves so dropping the field cannot quietly drop the information.

### 5. `forge_operator2flydsl` — done (`ad97421`)

### 6. Migration and validation

Registration and dynamic loading, prompt routing and the task-type prompt, shared Forge identity parsing, validator schema checks, and docs (`add-task`, `api-reference`, `what-is-aka`, plus the two places listing which task types need the flydsl package) are all updated, with coverage in `test_forge_operator2flydsl.py`, `test_sikl_tasks.py` and `test_task_validator.py`.

The GPU `task_validator` reports are the one item still open, and validating on MI355X/gfx950 turned up two things that change what those reports will say.

---

### Found on hardware: Arena cannot launch KernelForge as it currently ships (`c3ad1ed`)

Both Forge launchers resolved `kernel-agents`. KernelForge renamed that console script to `kernelforge`, kept the alias for one release and has since dropped it, on the stated reasoning that nothing in that repository, its docs or its examples invoked it. Arena did. Against Hyperloom at `e066d9a2e`, every `forge` and `forge_operator2flydsl` run died at launch before a task was read. Resolution now prefers the current name and still accepts the old one. Only the name changed — all 17 flags the rewrite launcher sends are options `forge-rewrite-by-flydsl` declares, rechecked against the installed package.

### Found on hardware: the production baseline does not clear the bundle's own bar (`bfe1ef1`)

Running the 21 tasks in their shipped state on MI355X/gfx950: 9 of the 17 GEMM tasks report `correctness: fail`, over **64 of their 221 workload points** — the same count as the acceptance run this PR quotes. At every one of those points aiter's own output fails identically, and a plain fp32-accumulating matmul passes the same gate on the same inputs, so the bar is reachable and what misses it is the tuned dispatch's accumulation. Two consecutive aiter calls on identical inputs also disagree beyond the tolerance at the small-M cases.

This matters for the validator gate you asked for: it runs a task's own `correctness_command` against the package as committed, where the editable source is a stub and the harness therefore grades the production implementation in the candidate's place. Left alone, 9 tasks would report `overall_status: FAIL` for a property the task deliberately does not gate on — the harness already treats the baseline's verdict as reported, never applied, and Arena never runs correctness before the agent. `bfe1ef1` adds an `operator2flydsl` family exception so that state is `SKIP/stub_candidate` rather than `FAIL`, conditioned on the run showing the baseline named as the implementation, a verdict for every declared case, and agreement between each case and its production-implementation line; performance must still run.

I have not changed anything about scoring, and I have not relaxed the correctness bar. Worth flagging separately: since the acceptance run applies these same callbacks, it currently rejects the production implementation at those points too.

### Also in the branch

`16fd3e8` and `329bc98` close a reward-hacking hole in the SIKL timing path. A case is timed over one set of input buffers, so an implementation could answer the first call and serve every replay from a cache keyed on their identity; correctness cannot see it, since it builds fresh inputs per case and calls once. The benchmark already supported being asked — `timed_run` exposes the invocation it measured and makes an unobservable capture fatal — and 16 task files elsewhere in the repository use it, so the suite now does too. The check is deliberately tolerance-free: it requires the replay to write over a NaN poison and to differ from the answer it gave before the inputs were redrawn, because any accuracy criterion here would reject the baseline for the reason above.

### End-to-end

All 21 tasks run compile, correctness and performance to a complete 13-case report, and all 21 reach KernelForge's OPTIMIZE phase through Arena on MI355X/gfx950. One caveat on that second number: the end-to-end run predates the two timing-gate commits, which were validated per-task rather than by re-running the full campaign set. 1055 tests pass; `make check-perf-helpers` passes.

PR #107, opened 2026-09-15
Unify task schema and agent evaluation for the SIKL interface

irvineoy:
This schema refactor adds compatibility with the **SIKL task interface**, supporting the **21 SIKL tasks introduced in #94**. These tasks need an explicit separation between the provided production baseline and an initially unimplemented FlyDSL candidate, task-specific workloads and numerical evaluation, and solution export through the shared framework.

To support SIKL while keeping tasks independent of agents, existing Arena tasks and the new SIKL tasks now use **one TaskSpec v2 schema in `config.yaml`**. All 438 retained tasks migrate to this common contract. Each task supplies its own inputs, reference, comparison rules and evaluation logic; Codex, Claude, Cursor, Forge and GEAK consume the same candidate and evaluator interfaces.

Initial task validation permits a declared empty candidate with a provided baseline. Final acceptance requires the submitted implementation to pass the task’s actual compilation, correctness and performance checks. The schema change therefore supports SIKL’s initialization and evaluation requirements while preserving a consistent interface for existing task families.

This branch builds on #94 (`329bc986`) and reconciles main at `0acf65b3`.

## Changes

- Consolidate `geak_v4` and `forge_operator2flydsl` into registry aliases for `geak` and `forge`. Move the unchanged GEAK Workflow runtime under `agents/geak/`; remove obsolete v1 launchers, duplicate defaults/docs and the unused Forge-specific exporter. Canonical and legacy names share launcher/configuration/post-processing.
- Require the Forge engine/lane Git root and tracked protected files before starting its native agent. Keep evaluation copies in a campaign-owned disposable directory, which the supervisor removes only after reaping descendants, including copies left by SIGKILLed drivers. Preserve logs, selected candidates and prior experiments. Copy-performance optimization and upstream capability/API changes remain separate follow-ups.
- Preserve final candidate acceptance and delivery status in text/CSV/JSON/dashboard reports, independently of measured scores and numerical correctness. Historical missing outcomes remain N/A. Completion queries now reject extra protected files without deleting them; actual evaluation retains its existing cleanup. Clarify that exporters read pre-export evaluation results, before final acceptance/delivery fields are written.
- Extend the task contract for the **SIKL interface** and migrate all **438 retained tasks across nine families**, including the 21 SIKL tasks, to the same `config.yaml` schema. Remove nine unmaintained repository tasks and their related configuration and documentation.
- Replace the old authoring guidance with [Task definition, schema, and authoring](https://github.com/AMD-AGI/AgentKernelArena/blob/9984622ff4d83cb63fcc130905aa0b957ebd119b/docs/how-to/add-task.md), required reading in `AGENTS.md` and `CONTRIBUTING.md`.
- Share task loading, contained path resolution, baseline snapshots, action/result parsing, workload coverage, scoring and exports. Tasks provide their own comparison rules. Diagnostic baseline results do not relax candidate correctness. Optional analysis tools remain evaluator-controlled.
- Update Codex, Claude, Cursor, native GEAK, task_validator and quality_loop to use the shared contract. Unify Forge initialization, translation and optimization under `forge`, keeping the rewrite name as a compatibility alias. Tasks need no agent-specific drivers.
- Update CLI defaults and flags, prompt transport, cleanup and native execution evidence. Current model defaults and dated checks are documented in [agent model defaults](https://github.com/AMD-AGI/AgentKernelArena/blob/97f19bb5dcdd1dc85944f342897a182f4db2d673/docs/reference/agent-model-defaults.md).
- Preserve original scored workloads and numerical gates while repairing references, empty-input handling, symbol boundaries and checks of actual timed outputs. Add task-specific negative controls and representative coverage.
- Fix GEAK’s launcher argument transport: the framework binds complete arguments in a per-run, hash-checked native dispatcher. The outer call supplies only the dispatcher path and an empty object. Actual child arguments, task rules, role budgets and native tool/result correlation remain unchanged.
- Qualify the proposed SGLang/ROCm 10 images on representative workloads. Retain the pinned SGLang 0.5.14/ROCm 7.2 MI355X default. [Runtime qualification](https://github.com/AMD-AGI/AgentKernelArena/blob/97f19bb5dcdd1dc85944f342897a182f4db2d673/docs/reference/runtime-upgrade-qualification.md) records the tested scope.

## Real agent execution coverage

**180/180 distinct `(agent, task)` combinations completed**: four agents, nine task families and five distinct tasks per family. **173 candidates passed independent final acceptance.**

| Agent | Completed / 45 | Accepted candidates |
| --- | ---: | ---: |
| Codex | 45 | 45 |
| Claude Code | 45 | 42 |
| Native KernelForge | 45 | 42 |
| GEAK | 45 | 44 |
| **Total** | **180 / 180** | **173** |

A completed bounded attempt can fail to produce a valid or faster candidate. Native status, candidate acceptance and speedup are recorded separately. Completed capability failures are retained rather than rerun for PASS. Failed startup, provider interruption, probes, repeated pairs and saved-candidate reevaluation do not add distinct completion credit.

The September 16 continuation uses Claude Code **Sonnet 5 / medium**, also as GEAK’s backend. It preserves each run’s actual model and source identity; the experimental GEAK Codex backend is excluded. Limits include one outer attempt, bounded runtime and GEAK search budget 2. Per-response output and search-budget limits are not an aggregate token or dollar cap. Claude’s 23-task continuation has a deduplicated CLI list-price estimate of USD 22.8283, not a subscription invoice; no complete GEAK billing total is claimed.

The seven unaccepted pairs remain visible: Claude’s unavailable FlyDSL API, MoE timing rejection and decode crash; Forge’s prohibited AITER delegation, partially unwritten MoE output and budget exhaustion without a candidate; and GEAK’s SIKL MoE candidate whose standalone correctness passed but all 13 performance-path numerical checks failed, followed by failed delivery and a missing public builder. The original Claude timing exception and decode crash cause, and GEAK’s underlying MoE numerical defect, were not fully recovered; the report distinguishes confirmed findings from these limits.

GEAK native role/schema failures remain failures even where independently evaluated candidate bytes pass. Matrix completion establishes bounded real task execution and recorded outcomes, not universal successful specialist search or speedup. The final Decode and Softmax candidates passed, while their native outcomes remained `FAILED`/`author_failed`.

One earlier decode attempt lacked the original correlated native terminal and receives no completion credit. A collector-only repair preserves raw native evidence before parsing model delivery; the single bounded successor recovered the required call/return evidence and completes that original pair once. Completed capability failures were not retried for acceptance. All campaign GPU/model jobs and launchers have ended; experiment artifacts remain preserved.

## Validation

- **Agent integration cleanup at `549a1a04`:** full Python 3.12 CPU suite: **14,108 passed, 6 subtests passed, 43 skipped** (optional dependencies/capabilities); **233 native-Forge CPU tests passed** in the pinned Docker environment, and `make check-docker-runner` passed. A real **MI355X / ROCm 7.2 RMSNorm bridge smoke passed all 10 cases** for baseline/candidate performance and candidate correctness; evaluation scratch was removed and diagnostics/candidates retained. Uses pinned Hyperloom `0425bde3`, makes **zero model calls**, and adds no optimization-pair or task-validator credit. SIGKILL of the supervisor itself or machine loss still requires explicit recovery. [Upstream interface follow-up](https://github.com/AMD-AGI/AgentKernelArena/pull/107#issuecomment-5705831986) records maintainer coordination; task numerical rules and Arena scheduling/scoring are unchanged.
- **Review follow-ups at `9984622f`:** 204 focused CPU regressions passed on Python 3.12, covering reporting, legacy compatibility, exporter lifecycle, completion queries, harness protection, evaluator, scoring, materialization and held-out behavior. Dashboard JavaScript syntax and Node rendering with a DOM stub passed. These fixes do not change task numerical gates, timing or score calculation; no GPU campaign was rerun for this follow-up. Final-head CI is reported separately in the PR checks.
- **438/438 task packages** have source-applicable, framework-finalized GPU validator PASS reports from Docker runs on compatible MI355X hardware using Codex medium. The aggregate reuses unchanged historical task versions and fresh reports for runtime-changed packages; it is **not** one fresh 438-task campaign on the final shared framework.
- The original main-merge revalidation’s **49 PASS / 1 WARN / 5 FAIL** outcomes remain intact. Repaired task versions passed later validation. One QKV task-level PASS retains a documented outer bytecode-inventory failure, with unchanged source bytes independently verified.
- **Core local CPU:** 14,019 tests and six subtests passed, six explicit skips, on immutable `f65e1e55`. **Runtime-source CI at `68ccbaac`:** both PR and push passed 13,972 tests and six subtests, with 149 explicit skips; helper audit, compilation and documentation passed. CI lacks optional native dependencies exercised locally. The final GEAK dispatcher change separately passed **230 focused CPU tests**, including actual JavaScript child-argument equivalence, and independent review. Overlapping counts are not additive.
- Five archived candidates affected by harness changes passed **35 fresh formal actions** using their exact saved bytes. These are not new searches or matrix pairs.
- Paired MoE timing completed **32 measurements / 512 case observations / 76 total public actions** on one MI355X. Both roles improved similarly after static host preparation moved outside timing; unchanged-candidate/baseline median latency ratios were 1.0007 and 1.0011. This is a harness-boundary study, not an agent gain.
- The quality-loop GPU smoke completed two eleven-case tasks, **28 passing actions, seven real Codex calls and two accepting independent reviews**. An out-of-allowlist enhancement was rejected before adoption. Untested promotion/enhancement branches are not claimed qualified.
- Real model-response/file-write smokes ran for Codex, Claude and Cursor. Runtime-upgrade, sanitizer, optional native dependency and historical evidence limitations remain explicit.

The [verification record](https://github.com/AMD-AGI/AgentKernelArena/blob/97f19bb5dcdd1dc85944f342897a182f4db2d673/docs/reference/verification-2026-09-15.md) records source-specific outcomes, all rejected candidates and the final matrix. The [main integration record](https://github.com/AMD-AGI/AgentKernelArena/blob/97f19bb5dcdd1dc85944f342897a182f4db2d673/docs/reference/pr107-main-integration.md) explains preserved task contracts and reconciliation decisions. Raw workspaces, logs, runtime checkouts and credentials remain outside Git.

Older GEAK results retain their original typed-completion evidence and lack full original tool-ID/session correlation; they do not retroactively qualify later strict transport paths. These results do not claim universal speedups, sanitizer coverage or whole-suite GPU qualification of every final framework/runtime combination.


jiaqiang-dot-liu:
Reviewed the `agents/forge/` path (I maintain KernelForge upstream). Two areas of feedback.

## 1. Monkeypatching KernelForge private interfaces

`upstream.install_hooks()` replaces 13 upstream symbols, pinned by SHA256 across 23 modules in `upstream_compatibility.json` (commit `0425bde3`, package `1.1.0`). Failing closed on an unreviewed release is the right default, but it hard-couples our release cadences.

Issues to fix:

- **`config.workspace` semantics under `rewrite` — please answer this one first.** `upstream.py:176-185` derives protected paths from `git ls-files` at `Path(values["config"].workspace)`, and `bound_candidate_root` assumes that path is the master engine root. If it is actually the attempt directory under `.forge_rewrite/`, that tree is gitignored, `git ls-files` returns nothing, and `extra_protected_paths` ends up empty — nested optimize sessions would run with no protected files at all. Confirming which it is determines whether the rewrite-path results stand.
- **`task_type="image_kernel"` hardcoded in 4 places** (`upstream.py:181`, `upstream.py:383`, `adapter.py:133`, `initialization.py:124`). All 438 tasks are labelled `image_kernel` to flip a multi-file switch. If `task_type` influences anything else upstream (prompt selection, expertise KB, backend heuristics, telemetry), per-family results are not interpretable. Better fix: add an explicit `multi_file` parameter upstream so Arena can pass the real task type.
- **`incumbent.py:33` hardcodes `len(measurements) != 3`.** If the benchmark sample count changes upstream, this returns `False` silently — incumbent recovery stops working with no error and no log. Everything else in this adapter fails loud; this one does not.
- **`runner.IterationLoop` is subclassed independently by `deadline.py`, `incumbent.py` and `round_budget.py`.** The resulting MRO semantics depend on the call order inside `install_hooks()`, which is not asserted anywhere. One composed subclass, or an explicit ordering check, would be safer.

Several of these patches are really API gaps on the KernelForge side. Happy to take these upstream:

- `protected_path_inventory` is O(n^2) — `protected_inventory.py` documents it consuming the model budget before the first call. That is a genuine upstream bug, not an Arena workaround.
- No total wall-clock bound on an implementer invocation; the outer Stop gate resumes with the original per-turn timeout (`deadline.bound_agent` works around this).
- Correctness/benchmark adjudication is not pluggable — the only seam is replacing `_run_canonical_suite`.
- `new_path_allowlist` supports single-directory globs only.
- `insession_gate` target/protected precedence is inverted when the implementation and its harness share a file.

With those landed upstream, `upstream.py` shrinks substantially and the hash pin can be relaxed.

## 2. Per-invocation workspace copies

`bridge.execute()` copies the whole workspace once per call (`bridge.py:64`); phases within one call share it. The unbounded caller is `arena_forge_driver.py`, so this is once per driver invocation, not once per task.

Using the cost model already in `round_budget.py:31-37` (the `* 5 / 3` factor): initial baseline measurement is 3 driver invocations, a KEEP assessment is 5 (3 bench + outer correctness + canonical), plus however many times the model runs the driver in-session. Roughly 6-15 copies per iteration, a few hundred to a couple thousand per campaign.

This is irrelevant for most tasks — a triton2triton workspace is a dozen small files. It only matters for image-backed/SIKL tasks, where `workspace.sources` materializes the whole `aiter_source/aiter` tree. I do not know that tree's size, so I cannot judge the impact; worth measuring before changing anything.

Two concrete items regardless:

- **Leaked copies.** `evaluation_workspace` relies on `TemporaryDirectory` cleanup, but `process_tree._stop_children()` sends `SIGKILL` (`process_tree.py:41`), which skips it. `artifact_root` is deliberately never cleaned (`adapter.py:161`) and the temp dirs live inside it, beside the task workspace. Every upstream stage timeout that kills an in-flight driver leaves a full workspace copy behind for the rest of the campaign. A SIGTERM grace period, or a campaign-level sweep list, would fix it.
- **`copy_workspace` walks the tree three times** (`bundles.py:38-55`): a symlink-escape pre-check, the copytree itself, and an absolute-symlink rebind pass. Two of those are pure validation of an immutable template and could be computed once per campaign.

If measurement shows the copy itself is significant, `cp --reflink=auto` on the same filesystem preserves the isolation the current code deliberately buys by avoiding hardlinks.

irvineoy:
@jiaqiang-dot-liu Following up on your review: agreed that KernelForge should own its engine behavior and expose the integration points Arena needs. Could you take the following changes in KernelForge, together with the corresponding Forge adapter wiring here?

1. Separate multi-file/editable-file support from `task_type`, so Arena no longer needs to label every task `image_kernel`. The task description and edit boundaries should be passed explicitly without selecting unrelated prompts or knowledge by accident.
2. Expose a single measurement-count contract for candidate acceptance and budget estimation. The Arena recovery path should consume it rather than hard-code three measurements (or a corresponding `5/3` cost estimate), and report a clear incompatibility if the contract changes.
3. Provide a stable way to compose the deadline, first-valid-incumbent recovery and per-round budget behavior. Currently these are Arena-side subclasses, so an upstream hook/API plus adapter wiring would let us remove the dependency on monkeypatch installation order. Merely changing the upstream loop would not by itself remove those adapter wrappers.

The boundary we want to preserve is: KernelForge owns search strategy; Arena owns task contracts, independent evaluation and run orchestration. Please keep the integration changes scoped to Forge, with no behavioral changes to Codex, Claude, Cursor or GEAK, and preserve Arena's worker queue, GPU allocation/masking, scheduling, invocation deadlines and cleanup. Task-owned cases/tolerances/timing, protected files, frozen baselines and final Arena scoring/export remain authoritative.

On the Arena side, we are adding a small engine/lane-root and protected-file inventory check and fixing cleanup of evaluation copies left by killed drivers. We are also consolidating the old Forge/GEAK names into registry aliases. Copy-performance optimization is deliberately deferred; we are not changing that policy or requesting another broad test campaign here.

irvineoy:
#109 has been merged into this PR's branch as `1a2b4933`. #107 remains open; this is an integration step, not an end-to-end validation sign-off.

The ordinary Forge GPU retest and ownership notes are recorded in https://github.com/AMD-AGI/AgentKernelArena/pull/109#issuecomment-5706118503 . The observations apply to #109 head `286eec00` integrated with the older #107 base `9984622f`; the current integrated revision has not received the same GPU retest.

Follow-ups to resolve/verify here:

- [ ] KernelForge: make Claude orchestration compatible with OAuth login (`--bare` currently ignores it).
- [ ] Forge/Arena integration: keep internal baseline/candidate timing comparable when task preparation replaces the driver; retain independent official task acceptance/scoring.
- [ ] Forge/Arena integration: expose native campaign failure separately from valid delivery of an unchanged initial candidate.
- [ ] KernelForge: make short-budget/zero-iteration outcomes clear; verify applicable default-lane contention and KB multi-file/case-contract boundaries before claiming support.
- [ ] Run a focused ordinary-task campaign with a real optimization attempt, delivery, and independent Arena evaluation on the resulting integrated revision.

These include existing upstream/integration issues and untested boundaries; they are not all demonstrated regressions caused by #109. KernelForge should retain ownership of search policy, with Arena responsible for its task contract, official evaluation, and faithful adapter reporting.

src/task_session.py
```diff
@@ -0,0 +1,433 @@
...
+        executed = self._execute("candidate", action, phase)
+        self.verify_candidate_harness()
+        if self._candidate_sources() != identity:
+            raise TaskExecutionError("Evaluation command modified candidate source")
+        return executed
+
+    def _candidate_sources(self, *, allow_missing: bool = False) -> dict[str, str]:
+        sources = {}
+        runtime_dirs = {"__pycache__", ".pytest_cache", ".git", "build"}
+        for edit in self.spec.candidate.editable:
```
irvineoy, line 406: [P1] Follow-up: bind completion evidence to all candidate implementation dependencies

Deferred for future work. `_candidate_sources()` fingerprints only `candidate.editable`, while the harness guard permits newly created files whose names do not match its protected-file patterns. A declared `kernel.py` can import a newly added `helper.py`, pass evaluation, and then have its behavior changed by editing only `helper.py` without invalidating the saved completion. In a CPU end-to-end fixture, changing the helper's returned value from 3 to 99999 after evaluation still left `task_run_is_complete(...) == True`; the recorded candidate sources contained only `kernel.py`. Resume can consequently reuse correctness/timing evidence for a different implementation. Reject undeclared implementation dependencies or include every allowed dependency in source/completion identity, with a regression that changes an imported helper after acceptance.

src/task_execution.py
```diff
@@ -0,0 +1,136 @@
...
+        try:
+            stdout, stderr = process.communicate(timeout=timeout)
+        except BaseException as exc:
+            # Kill the command's process group, including timed-out compiler/JIT
+            # descendants. They must not keep consuming a later action's GPU.
+            try:
+                os.killpg(process.pid, signal.SIGKILL)
+            except ProcessLookupError:
+                pass
+            stdout, stderr = process.communicate()
```
irvineoy, line 58: [P2] Follow-up: bound timeout cleanup when descendants retain output pipes

Deferred for future work. After a timeout, `killpg(process.pid, SIGKILL)` does not reach a descendant that started a new session. If that descendant inherited stdout/stderr, the subsequent unbounded `communicate()` waits for its pipe handles to close, so the declared action timeout no longer bounds execution. A CPU reproduction with a 0.2-second action timeout and a detached child sleeping for 3 seconds returned only after about 3.03 seconds; a child that never exits can block the worker indefinitely. Track and clean up the owned descendant tree and bound the final pipe drain. Add coverage for a detached descendant retaining stdout/stderr, while preserving unrelated processes.

src/held_out/evaluation_v2.py
```diff
@@ -0,0 +1,221 @@
...
+                if source.is_file() and not set(source.relative_to(workspace).parts[:-1]) & {
+                    "__pycache__", ".pytest_cache", ".git", "build",
+                }:
+                    result[relative] = hashlib.sha256(source.read_bytes()).hexdigest()
+    return result
+
+
+def _case_input(row: dict) -> str:
+    # Renaming an existing case does not make its input held out.
+    return json.dumps({key: row.get(key) for key in ("shape", "dtype", "params")}, sort_keys=True)
```
irvineoy, line 70: [P2] Follow-up: exclude case identifiers from held-out input identity

Deferred for future work. `_case_input()` treats all of `params` as semantic input, but SIKL's `scripts/task_contract.py` includes the case `uuid` in that dictionary. SIKL input construction uses dimensions and seed, not the UUID. Changing only the 13 UUIDs in the GEMM workload, while keeping dimensions/axes/seed unchanged, makes every original case evade the overlap check. A CPU end-to-end held-out fixture also accepted a UUID-only injection and produced `generalization_status=both_pass` with a nonzero score (fixture timings were synthetic). This can report performance on already-seen inputs as generalization. Separate identifier metadata from task-declared semantic input identity and add a regression requiring UUID-only changes to be rejected.

agents/forge/adapter.py
```diff
@@ -0,0 +1,466 @@
...
+                    # and harness, which is what this value means upstream. Do
+                    # not drop it for being redundant with --source-files: that
+                    # only resolves multi-file handling, and a single-file
+                    # candidate would resolve to no task type at all, losing the
+                    # colocated test/reference protection and the entry-point
+                    # prompt sections with it.
+                    "--task-type", "repository",
+                    "--baseline-json", plan["baseline"], "--program-md-file", plan["program"],
+                    "--agent-backend", config["agent_backend"],
+                    "--session-timeout-sec", str(config["session_timeout_seconds"])]
```
irvineoy, line 260: [P2] Follow-up: disable automatic preparation of Arena's driver for ordinary optimize

Deferred for future work. This `forge-loop` command leaves `--prepare-task` enabled. In the inspected KernelForge implementation (Hyperloom commit `4473e7041ddb9fbead88541d7d4ab6953444bd06`), the fresh-campaign preflight requires profiling and invokes the preparation agent to author/repair the driver if preflight fails. Arena's generated bridge deliberately returns exit code 2 for `--profile-run`; executing the upstream profiling check against the generated driver reproduced that failure on CPU. KernelForge then binds its canonical driver digest to the post-preparation bytes, so that integrity check does not preserve Arena's original measurement entrypoint. Search may use altered correctness/timing logic and spend its budget selecting a candidate rejected by Arena's independent final evaluation; this does not demonstrate a bypass of final Arena scoring. The rewrite command already passes `--no-prepare-driver`, and its nested loop passes `--no-prepare-task`, so rewrite validation does not cover ordinary optimize. Pass `--no-prepare-task` for ordinary optimize, handle unsupported profiling without rewriting the trusted bridge, and cover the real optimize preflight path. No GPU or live preparation-agent run was performed for this reproduction.

irvineoy (commented):
Recording the four agreed follow-up issues for future reference against e20573efaca8c047c4079341ea8233060dc44d6e. Per the requested disposition, these are deferred and this review does not request changes before merging. They remain unresolved; merging does not constitute a fix. Reproductions described below are CPU checks, not GPU validation.

PR #109, opened 2026-09-16
Let the Forge engine own its search policy

jiaqiang-dot-liu:
## Summary

The Forge adapter overrode six engine search decisions on every campaign. This
removes them so Arena publishes the task and evaluates the delivered candidate,
and leaves the search policy to the engine.

- `agents/forge/adapter.py`: drop `--no-rewrite-kb` (rewrite path) and
  `--no-experience-kb`, `--no-prepare-task`, `--no-profiling`,
  `--no-specialist-probe`, `--lanes 1` (forge-loop path).
- `agents/forge/upstream.py`: `configure_nested_loop` no longer appends
  `--lanes 1 --no-profiling --no-specialist-probe` to the nested loop.
- `--no-prepare-driver` is deliberately kept on the rewrite path. It does not
  skip verification: upstream runs `preflight_rewrite_driver` unconditionally
  and the flag only gates the repair step, so keeping it means a non-conforming
  driver fails honestly instead of being rewritten by an agent.

## Measured effect of re-enabling the knowledge base

`SIKL-task/gemm_a16w16_nt_n32_k6144`, one MI355X (gfx950, ROCm 7.2), Claude
Opus 5, Hyperloom `0425bde3` source checkout. Two runs differing only in this
change:

| | before | after |
| --- | --- | --- |
| PORT attempts | 1 | 0 (skipped) |
| time to first candidate | 789 s | 80 s |
| candidate latency | 0.0216 ms | 0.00562 ms |
| vs AITER baseline (0.0072 ms) | 0.329x | 1.356x |

The warm start is not a blind reuse. It read the published candidates for
`kernel:flydsl:gemm_a16w16_nt_n32_k6144:aiter:unspecified:flydsl:mi355x`, ran
each through this task's own driver, rejected three that failed to compile and
adopted the fastest survivor after all 13 cases passed.

## What is not validated

- The `--no-prepare-task` removal is **unverified**. It only affects the
  optimize/initialize paths, which no SIKL task uses. Upstream documents the
  profiling interface as mandatory under `--prepare-task`, and this adapter's
  bridge answers `--profile-run` with an explicit unsupported capability that
  `loop/task_preparer.py` does not parse. A forge-loop campaign may therefore
  fail preflight or invoke the prep agent to rewrite `arena_forge_driver.py`.
  A `flydsl2flydsl` run should confirm the behaviour before this lands.
- `--lanes` returning to the upstream default of 3 has not been measured for
  GPU contention on a single device.
- Warm starts publish into the shared default producer index. Candidates scored
  against a TaskSpec v2 case set now share that index with native runs whose
  scoring boundary differs.
- A single-file upstream recipe cannot describe a multi-file candidate, so a
  task whose `candidate.editable` spans more than one file may receive a warm
  start covering only part of its submission. That was the original reason for
  disabling warm starts; gating on the editable file count would be a narrower
  fix than the blanket override removed here.

## Test plan

- [x] `tests/test_forge_v2.py`, `test_forge_backend_coverage.py`,
      `test_forge_operator2flydsl.py`, `test_forge_action_budget.py`,
      `test_forge_initialization.py` — 91 passed, 16 skipped
- [x] `agents/forge/upstream.py --arena-probe` against the pinned checkout
- [x] Two full GPU campaigns on the task above, cold and warm
- [ ] A `flydsl2flydsl` campaign exercising the forge-loop path with task
      preparation enabled


irvineoy:
Thanks for moving search-policy ownership back to KernelForge. We are integrating this into the #107 branch and will address the remaining integration issues there before deciding whether #107 is ready for main. This is not a claim that the ordinary optimization campaign passed end to end.

I tested #109 head `286eec00` integrated with the then-current #107 base `9984622f`, using one MI355X, ROCm 7.2, Hyperloom `0425bde3`, Claude Code 2.1.273, Sonnet 5 with medium reasoning, and `flydsl2flydsl/rmsnorm_kernel` with all 10 cases. The subsequent #107 changes have not received this same GPU retest.

**Observed results and follow-ups:**

1. **Automatic preparation works, but changes the internal evaluation path.** It replaced `arena_forge_driver.py` (the bridge to public task actions) with its own correctness/graph-timing implementation and added `graph_harness.py`. Native preflight passed. The unchanged kernel then measured about 0.014 ms internally versus the Arena baseline of about 0.00265 ms, starting native search near 0.187x. This is a measurement-protocol discrepancy, not evidence that the kernel became slower. Please keep internal baseline/candidate comparisons consistent: either retain the task actions or clearly separate an internal proxy with its own matching baseline. Arena's final independent evaluation remained on the original task harness. **Ownership: shared Forge/Arena integration contract.**

2. **Claude OAuth works in preparation/analysis but fails in orchestration.** With a 60-minute campaign cap, orchestration failed three times with `Not logged in`, completing zero candidate iterations. The pinned upstream defaults `AgentToolPolicy.bare=True`, and this orchestration path adds `--bare`. A same-login A/B probe succeeded normally and failed with `--bare`; that mode skips OAuth credentials. Please make the mode compatible with the selected authentication method and cover the different agent roles. This is not established as a regression introduced by #109; an API-key setup was not tested. **Ownership: KernelForge's Claude backend.**

3. **Run failure and valid candidate delivery need separate reporting.** Despite native `orchestration_failed`, the engine returned exit code 0; the Arena adapter selected the initial validated implementation and reported `DELIVERED`. Arena accepted that unchanged kernel at approximately 1.00x; its source hash was identical to the original. That correctness result is valid, but it does not establish a healthy optimization campaign. Please propagate the native termination reason and expose candidate validity separately from campaign health. **Ownership: KernelForge result contract plus Arena adapter/reporting.**

4. **Short-budget behavior needs to remain visible.** A separate 30-minute run completed preparation/analysis but ended with `round_budget_exhausted` and zero optimization iterations. Let Forge own admission/search policy, while making setup cost and zero-iteration outcomes visible. Default multi-lane GPU contention was not exercised because the longer run stopped at orchestration. **Ownership: primarily KernelForge; Arena exposes the result.**

5. **Keep the acknowledged KB boundaries tracked.** The tested SIKL warm start is useful evidence. Multi-file submissions still need complete restoration (or an explicit unsupported path), and history measured against different case sets/timing contracts must be revalidated before comparison. These are outstanding coverage gaps, not failures reproduced by this test. **Ownership: primarily KernelForge KB, with Arena supplying the submission/task contract.**

A focused follow-up on #107 should exercise an actual optimization attempt, candidate delivery, and independent Arena evaluation after these fixes. No need to repeat the entire 180-run matrix just for this change. The intended boundary remains: KernelForge owns how it searches; tasks/Arena own official acceptance and scoring, and the adapter must faithfully carry results between them.

PR #110, opened 2026-09-17
Drive Forge through the engine's published interface

jiaqiang-dot-liu:
Six commits on top of `codex/task-schema-refactor`, net **+743 / −2737** lines: twelve adapter modules and six test files are removed, and the adapter now talks to KernelForge over its command line and measurement driver only.

## Problems solved

**The adapter patched the engine instead of calling it.** `install_hooks` replaced twenty KernelForge symbols at run time, most of them private: the canonical correctness gate, the seed writer, the PORT prompt builder, the git commit helper, the nested-loop argv, the apply-back patch generator, and three independent subclasses each of `IterationLoop` and `InSessionGate` whose resulting MRO depended on install order. To keep those patches honest the adapter pinned twenty-three upstream modules by SHA256, which bound the two release cadences together byte for byte — a new engine release could not be adopted without regenerating hashes. Removed the hook installer, the compatibility pin, and the eight modules that existed only to patch (`protected_inventory`, `gate_targets`, `deadline`, `incumbent`, `round_budget`, `codex_auth`, `port_budget`, `action_budget`). `upstream.py` then only forwarded argv, so rewrite and optimization campaigns now exec `kernelforge.cli` directly; `engine.py` keeps the two jobs that genuinely need the engine importable and refuses anything else.

**Every assessment copied 1.3 GB.** Each bridge invocation materialized a private tree from the task package so a candidate could not reach the reference, harness, or a concurrently measured baseline. For an image-backed task that is the whole declared source tree — the SIKL GEMM workspace is 1.3 GB across 1264 files, of which 17 belong to the task and the rest is the AITER tree the runtime only reads. One campaign copied it about ten times before its first iteration, to guard numbers nothing downstream trusts: a mid-search measurement is the engine's own signal, and the framework re-runs every candidate action in its own workspace before deciding the verdict. Candidate actions now run in the engine tree; baseline actions keep a private tree built from the framework's frozen snapshot, so a candidate that breaks the runner cannot also move the anchor it is measured against.

**The engine was handed two budgets.** The adapter passed both `--max-hours` and `--deadline-unix`; given an absolute deadline the engine ignores the relative one, and given a relative one it dates the deadline from its own process start. That start trails argv construction by interpreter startup, the capability probe and the engine import, measured at 42s on MI355X. It now receives one relative budget behind a ten-minute margin, with the campaign's hard timeout as the backstop rather than the first thing to fire.

**A campaign could build a candidate and report nothing.** Two paths discarded work the engine had already committed. The rewrite CLI folds framework apply-back into its own `success`, so it exits nonzero whenever that patch fails; the adapter read that exit code as an engine failure. One MI355X run scored a correct **1.361x** FlyDSL candidate as **0** because the apply-back agent could not start. Arena requests no framework patch and cannot decline the stage — the engine infers it from the campaign workspace having a resolvable Git HEAD, which its agent sessions require before any port attempt. A rewrite whose result reports a completed port, no failure class, and that patch as its only failed stage now keeps its outcome. Separately, a timeout discarded everything including a search that had earned KEEPs; recovery now reads `best_result.json` (the engine publishes each KEEP before it may start another session, and defers termination signals across that publication), then `run_state.json`'s `head_commit`, then the commit Arena itself accepted before launch. Delivery also stops falling back to `best_commit`, which names the apply-back commit once that stage ran and whose tree carries framework edits rather than the attempt's bundle.

**The README documented code that no longer existed.** It still described `upstream.py`, the hash pin, the run-time symbol replacements, a `port_budget.json` and an `evaluation-workspaces/` directory that nothing writes, a `codex_auth_mode` setting that went with its implementation, and an adapter-owned round-admission budget. Rewritten against the code that ships, including the regression list the patch removal left in a commit message where no adapter reader would find it.

## Verification

`python3 -m pytest -q` over the nine test files that import `agents.forge`: **206 passed, 9 skipped**. New behaviour is covered by regression tests that were confirmed to fail without the change (8 failures on the previous adapter). Seven collection errors elsewhere in `tests/` are pre-existing and reproduce on a clean tree.

Evidence for the delivery fixes comes from two real MI355X/gfx950 campaigns on `SIKL-task/gemm_a16w16_nt_n32_k6144`: one where apply-back failed and a valid 1.361x candidate scored 0, and one where it succeeded and Arena accepted at 1.37x (score 257.1).

A third campaign on the same task exercises this adapter against AMD-AGI/Hyperloom#1596, which supplies the engine-side changes the review asked for. Apply-back was declined (`applyback_reserve=0s`, `applyback_required: false`). No KB warm start was taken: the one stored entry was rejected as `correctness_failed`, so PORT ran cold and passed on attempt 1 at **1.128x**. OPTIMIZE iteration 1 then produced a genuine KEEP at **1.224566x** (commit `fbcbdf747169`), which Arena re-measured and **accepted**: score **244.78**, speedup **1.2478x**, delivery **COMPLETE**, `failure_class: ""`, `terminated_for_deadline: false`. This is the first campaign on hand reaching `cumulative.kept: 1`, and the first to show a v2 task clearing the final KEEP rather than only delivering a PORT.

## Known gaps

Search behaviour stays behind what the removed patch layer forced until the engine grows the seams it took: no injected protected paths, no total wall clock on an implementer invocation, no pluggable correctness adjudication, and upstream's own seed and PORT prompt. The apply-back gap is closed where the engine cooperates: 89b8525 asks an engine reporting `applyback_optional` to skip the stage, and AMD-AGI/Hyperloom#1596 defines that option, so the 20-minute reserve returns to the search. Against an older engine the stage is still attempted, still holds that reserve, and still edits the engine's disposable copy of any framework sources the task materialized.

**The timeout-recovery path has still not been exercised on GPU**, though its first tier is no longer hypothetical. The cold-PORT campaign above reached `cumulative.kept: 1` and wrote a real `best_result.json`, so the artifact recovery reads first now exists in a live workspace. Recovery itself — killing a campaign at the deadline and delivering from that file — remains covered only by the engine's publication contract and CPU tests.

**The multi-file candidate commit has no GPU evidence.** The KEEP commit above contains a single `kernel.py`, because the attempt directory held only that file. AMD-AGI/Hyperloom#1596 commits the whole attempt directory and covers the helper case in a CPU test against the real `_ensure_git_committed()`, but no campaign so far has produced a helper alongside the main kernel.

## Test plan

- [x] `pytest` over all `agents.forge` dependents on CPU
- [x] Regression tests verified to fail without the adapter change
- [x] Real MI355X campaigns for the apply-back and delivery paths
- [ ] A GPU campaign that earns a KEEP and is then killed, to exercise timeout recovery end to end
- [x] Cold-PORT E2E run on MI355X: KEEP at 1.224566x, Arena accepted at 1.2478x (score 244.78)
- [ ] A campaign producing a declared helper beside the main kernel, to cover the multi-file commit on GPU


irvineoy (commented):
Two reproduced integration regressions need upstream KernelForge interfaces plus the corresponding Arena wiring before these compatibility hooks can be removed. Both reproductions are CPU-only; they do not claim GPU end-to-end validation.

agents/forge/upstream.py
```diff
@@ -1,447 +0,0 @@
...
-            stdout, stderr = await communicate_process_group(process, timeout=timeout_cap_sec)
-        except asyncio.TimeoutError:
-            return modules.canonical_correctness.CanonicalCorrectnessResult(
-                passed=False, detail="Arena v2 correctness timed out", outcome="timeout")
-        output = (stdout + stderr).decode(errors="replace")
-        return modules.canonical_correctness.CanonicalCorrectnessResult(
-            passed=process.returncode == 0,
-            detail="Arena v2 compile + correctness", output=output[-4000:])
-
-    modules.canonical_correctness._run_canonical_suite = canonical
```
irvineoy, line 277: [P1] Provide the upstream v2 acceptance interface before removing this bridge

Against PR head `775b6301` and the previously qualified Hyperloom `0425bde3`, removing this override restores `kernelforge.loop.canonical_correctness._load_suite()`, which still reads top-level `compile_command` / `correctness_command`. The v2 task packages instead declare `evaluation.runner`. Once a candidate clears the performance threshold, `IterationLoop` calls `accept_candidate()` and gets a missing-command failure, so the candidate receives `kept=False` even though the Arena driver accepted it.

I reproduced this on CPU with this PR's `fixture_task`: `bridge.execute(..., role="candidate", action="correctness").passed` returns `True`, then the real upstream `accept_candidate(engine_root, ...)` returns `passed=False` with `config.yaml declares no 'compile_command'`. This exercises the actual upstream acceptance function, without a model or GPU. The reported GPU delivery/speedup does not establish that this path works: a PORT candidate can be delivered without a subsequent KEEP, and the PR currently reports zero observed KEEPs.

Please add a public final-acceptance command/callback in Hyperloom and wire Arena's v2 driver into it before removing the compatibility bridge. Keep nonzero exits and timeouts as failures; bypassing the final check would not fix the task contract. Add a CPU integration regression using the real upstream acceptance function and a v2 task (including a rejected candidate), rather than a mocked engine result. An unsupported installed engine should be rejected during capability preflight.

agents/forge/upstream.py
```diff
@@ -1,447 +0,0 @@
...
-    modules.port_loop.run_port_loop = bound_port(modules.port_loop.run_port_loop, plan)
-    modules.runner.run_port_loop = modules.port_loop.run_port_loop
-    original_commit = modules.runner._ensure_git_committed
-
-    def commit(workspace, message, paths, *, branch=""):
-        root = bound_candidate_root(plan, Path(workspace))
-        if message == "forge-rewrite: initial correct flydsl port":
-            allow_candidate_paths(Path(workspace), spec, prefix=root.relative_to(workspace).as_posix())
-            paths = [*paths, str(Path(workspace) / ".gitignore")]
-        paths = list(dict.fromkeys([*paths, *(str(path) for path in candidate_files(spec, root, required=False).values())]))
```
irvineoy, line 330: [P1] Preserve complete candidate commits for multi-file rewrite tasks

This deleted wrapper adds all declared candidate files to the PORT commit. In Hyperloom `0425bde3`, `rewrite_by_flydsl.runner` only passes `[spec.flydsl_kernel]` to `_ensure_git_committed()` after PORT. For a candidate whose helper is present on disk but not already staged/committed, the driver can validate the complete implementation while the selected commit contains only the main kernel. Arena correctly extracts delivery from that immutable commit, so the helper is then missing.

CPU reproduction using the PR's two-file fixture and the real upstream `_ensure_git_committed()`:

```text
PORT_CANDIDATE_CHECK=True
ATTEMPT_COMMITTED_FILES=.forge_rewrite/current/source/kernel.py
DELIVERY_FAILURE=Missing candidate file: source/helper.py
```

The existing complete-bundle launcher test misses this because `mock_engine()` uses `git add -f .forge_rewrite/current`, committing the entire directory; the real PORT path only adds the anchor. The reported GPU task also declares only one editable `kernel.py`.

Please add explicit candidate-bundle/editable-path support in Hyperloom and pass the task declarations from the Arena adapter, so PORT and subsequent KEEP commits preserve the complete validated candidate, including declared helpers and deletions. Do not blanket-add protected workspace files or fill missing files from the mutable working tree at delivery time. Add an integration regression using the real upstream commit behavior and a declared main file plus helper.

agents/forge/upstream.py
jiaqiang-dot-liu: Addressed upstream in AMD-AGI/Hyperloom#1596 (`6d015e78`), but with a different design than you asked for, so please read this as a proposal rather than a claim that the thread is closed.

**What changed.** `IterationLoop` now calls `accept_candidate()` only for the assembly backend:

```python
# Step 7: the numerical contract, which only assembly declares and only it needs.
if improved and self.ic.kernel_backend == "assembly":
```

Rather than adding a second, pluggable acceptance door, the change removes the one that duplicated the first. Your reproduction is exactly right: `_load_suite()` reads top-level `compile_command` / `correctness_command`, a v2 package declares `evaluation.runner`, and the KEEP dies there.

**Why removal rather than a new interface.** By Step 7 the candidate has already been compiled and checked by the driver in Step 4, and every measurement since came through that same driver. `accept_candidate()` re-asked that question through a config-shaped door, which required the engine to know a task-configuration shape that is not its business. Two authorities for one question, and the second is the one that cannot read a v2 task.

**Where your objection still bites.** You wrote that bypassing the final check would not fix the task contract. For a caller-supplied driver I believe this is deduplication rather than a bypass, because the driver is the task's own authority and Arena authors it. It is weaker where the engine prepares the driver itself (`--prepare-task`), since the producer then influences its own judge. That case is not Arena's, but KernelForge has other callers, and if you think it needs a separate guard I would rather hear it before the engine ships this.

**Two of your requirements are not met yet, and I would rather say so than let them pass.**

1. Capability preflight. #1596 adds `applyback_optional` to `capabilities()` but exposes nothing for this change. Against an engine without #1596, every candidate is still rejected at acceptance, `cumulative.kept` stays 0, and nothing fails: the campaign completes, delivers the PORT bundle and reports a speedup. That is the exact failure mode that misled the earlier runs. I will add a capability flag upstream and have the Arena probe fail preflight when it is absent.

2. A rejected-candidate regression. #1596 deleted `test_a_warm_start_failing_the_task_suite_is_not_adopted_or_published` and did not replace it. Under the new design a test against the acceptance function is moot, but the equivalent property is not covered: a v2-shaped task whose driver fails must not produce a KEEP. I will add it.

**GPU status.** Not established yet. A cold-PORT campaign on `SIKL-task/gemm_a16w16_nt_n32_k6144` is running now against this engine with no KB warm start, so PORT and OPTIMIZE both run for real. No KEEP so far. I will post the outcome either way rather than infer it from a PORT delivery, which as you noted proves nothing about this path.

agents/forge/upstream.py
jiaqiang-dot-liu: Fixed upstream in AMD-AGI/Hyperloom#1596 (`37139c02`). PORT now commits the attempt directory instead of the anchor file:

```python
_ensure_git_committed(
    workspace,
    "forge-rewrite: initial correct flydsl port",
    [attempt.relative_root],
    branch=optimize_git_branch,
)
```

**The mechanism differs from what you asked for.** No candidate declarations are passed from Arena. The attempt directory is already the candidate boundary: the engine creates it, only the session writes into it, and nothing else in the workspace lives there. Committing it records every file the driver validated without the engine needing to know the task's declaration schema.

Passing the declarations would have required widening the SIKL `editable` scope from a single file to a tree, and that scope also bounds what the agent may edit. Widening it to carry helper files would have opened `config.yaml`, `workload.json` and the materialized `aiter_source/` to the session. That looked like the worse trade against letting the producer's own directory define the bundle.

**Your two prohibitions hold.** Nothing outside the attempt directory is staged, so no protected workspace file is blanket-added. Delivery still reads the immutable commit; nothing is filled in from the working tree.

**Deletions are covered.** `git add -- <dir>` stages removals, so a helper the port deletes is recorded as deleted rather than silently surviving in the commit. I verified that rather than assuming it.

`__pycache__` is excluded at every depth, including directories nested below the attempt root, because the interpreter writes caches beside any module the driver imports.

**Coverage.** `test_ensure_git_committed_takes_a_whole_attempt_directory` exercises the real `_ensure_git_committed()` with a main file, a helper and a nested subdirectory. It lives upstream, not here. Your point about `mock_engine()` stands: its `git add -f .forge_rewrite/current` now happens to match what the real PORT path does, which is coincidence rather than coverage. I will add an Arena-side regression against the real upstream commit behaviour.

Both threads depend on #1596 landing; this PR as it stands assumes that engine.

agents/forge/upstream.py
jiaqiang-dot-liu: **GPU evidence for this thread (MI355X / gfx950).**

Cold-PORT campaign on `SIKL-task/gemm_a16w16_nt_n32_k6144` against Hyperloom #1596 + this PR's adapter (`--no-applyback`, Arena-authored driver with `--no-prepare-driver`):

- PORT OK on attempt 1 → interim **1.128x**.
- OPTIMIZE Iteration 1 produced a real KEEP: **mean case speedup 1.224566x** (`+8.6%` vs the PORT incumbent), commit `fbcbdf747169`. In-session gate scores were `[1.247389, 1.248487, 1.177824]` before the KEEP.
- The engine then published that KEEP to the remote KB (`read_reason` for the prior warm-start entry was `candidates_rejected` / `correctness_failed`; the new write succeeded).
- Arena re-measured and **Accepted** the delivery: score **244.8**, speedup **1.25x**, Delivery **COMPLETE**. Engine result: `success: true`, `failure_class: ""`, `applyback_required: false`.

This is the path your reproduction predicted would die at `accept_candidate()` with `config.yaml declares no 'compile_command'`. It did not. A PORT-only delivery is no longer the only thing we can show; the KEEP itself landed.

What this still does **not** close: the capability-preflight flag and the rejected-candidate regression I named in the previous reply. Those remain open work items, independent of this GPU result.

agents/forge/upstream.py
jiaqiang-dot-liu: **GPU evidence for this thread (same campaign as the other reply).**

The KEEP commit `fbcbdf747169` records:

```
.forge_rewrite/20260921-114031-22d585fb/kernel.py
```

That matches the attempt directory on disk at KEEP time: only `kernel.py` existed there. The PORT commit `3024470` was the same shape. So this run confirms the directory-commit path does not drop the validated single-file candidate, and Arena's delivery from that immutable commit was COMPLETE.

It does **not** exercise the multi-file failure you reproduced (helper present on disk, driver passes, commit contains only the anchor). No helper was produced in this campaign. The Arena-side regression against real upstream `_ensure_git_committed()` that I owed in the previous reply is still outstanding; the CPU test in Hyperloom #1596 (`test_ensure_git_committed_takes_a_whole_attempt_directory`) remains the only coverage of the helper case.

jiaqiang-dot-liu:
## E2E verification (MI355X / gfx950)

Cold-PORT run of `SIKL-task/gemm_a16w16_nt_n32_k6144` against this PR's forge adapter and Hyperloom #1596 (`feat/rewrite-caller-contract`), with remote KB enabled and `--no-applyback`.

| Stage | Result |
| --- | --- |
| Apply-back | Declined (`applyback_reserve=0s`, `applyback_required: false`) |
| KB warm-start | Read a prior 1.1913x entry; rejected (`correctness_failed`); cold PORT |
| PORT | OK attempt 1 → interim **1.128x** |
| OPTIMIZE KEEP | Iteration 1 → **1.224566x** (commit `fbcbdf747169`), published to KB |
| Arena verdict | **Accepted**, score **244.8**, speedup **1.25x**, Delivery **COMPLETE** |

Replies on the two P1 threads spell out what this does and does not prove relative to each comment.

irvineoy:
Recording two **non-blocking follow-ups for future reference** before merging this PR into #107. These remain pending; merging does not mean they are fixed. This assumes Arena is used with the updated Hyperloom, including AMD-AGI/Hyperloom#1596; compatibility with older engines is outside this decision.

1. **Preserve newly added helpers in subsequent KEEP commits.**

   The initial PORT now commits the complete attempt directory, which fixes the original PORT omission. Subsequent optimization uses a different commit path: the nested optimize invocation does not supply the candidate's new-file scope, and the ignored `.forge_rewrite/` tree is not discovered by the ordinary untracked-file scan. A later iteration can therefore pass the driver with a new helper present on disk, but save a KEEP commit without that helper. Delivery correctly reads the immutable commit and then lacks a required file. This was reproduced on CPU using the real upstream commit behavior with Hyperloom `4473e704`, which includes #1596.

   Follow-up: carry the allowed candidate-file scope through initialization and later optimization, preserving additions, modifications, and deletions when saving or discarding candidates. Add a regression that creates a helper after PORT, performs a KEEP, extracts that commit, and checks the delivered bundle. Do not fill missing files from the current working directory: those files may belong to a later, unvalidated iteration.

2. **Keep the Arena-supplied driver unchanged in ordinary optimize.**

   The ordinary optimize command still leaves Hyperloom's `--prepare-task` default enabled. Preparation requires profiling support, while the Arena bridge explicitly rejects `--profile-run`; this can start an agent that rewrites the supplied driver. Search-time correctness/timing can then differ from Arena's intended protocol. Arena still independently evaluates the delivered code, but the search may waste budget following misleading feedback.

   Follow-up: disable automatic preparation for the Arena-supplied driver and ensure unsupported profiling is disabled or capability-gated. A reusable profiling entrypoint can be added separately. Add coverage for ordinary optimize that checks the driver remains unchanged. Rewrite already passes `--no-prepare-driver`, and its nested optimize disables task preparation, so the successful rewrite run does not cover this ordinary-optimize behavior.

The reported MI355X run demonstrates a real KEEP and accepted delivery for a **single-file rewrite**. It does not exercise a helper added during subsequent optimization or the ordinary-optimize preparation path. We are accepting these two limitations as follow-up work rather than blocking this merge.

